PR #84, opened 2014-04-25
Compatibility with active record 3.0 and 3.1

sobrinho:
Calling `super()` makes the active record adapter compatible with older versions of active record.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,8 @@ jdk:
 
 env:
   - TEST_BUNDLE_GEMFILE=$PWD/elasticsearch/elasticsearch-rails/elasticsearch-model/gemfiles/3.0.gemfile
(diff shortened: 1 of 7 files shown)


dre-hh:
@karmi we've got rails 4.2 by the time now. I've personally upgraded some rails projects from 3.0 to 3.2. I don't remember wether there was some minor incompatibilities in AR. The whole process was actually pretty painless. I suggest we close this issue and encourage people to upgrade at least to AR 3.2

karmi:
@dre-hh Agreed that Rails 3.0 is a legacy thing now, however, the gem works with that version, so I'm not sure if we should remove it... On the other hand, I'm all for adding support or 3.2.x and getting rid of 3.0.x _if it would complicate things_.

sobrinho:
Closing, we finally upgraded the project, thanks!

karmi:
@sobrinho, that's the best outcome! :) Best of luck!

PR #110, opened 2014-05-08
Coerce pagination string params

ryansch:
We hit this in our app when we pass in params from a rails controller.  This PR causes the `limit` and `offset` methods to work like the `page` (or `::Kaminari.config.page_method_name`) method already does.


ryansch:
I've signed the CLA.

karmi:
@ryansch Finally got to merge this, sorry for the ridiculous delay!

PR #113, opened 2014-05-09
Allow custom transformations to be defined for bulk imports.

AaronRustad:
This allows for the customization of bulk imports. This was driven out of need to include the parent in the index payload, though could extend to any bulk api options.

``` ruby
transform = lambda do |article|
  {index: {_id: article.id, _parent: article.author_id, data: article.__elasticsearch__.as_indexed_json}}
end

Article.import transform: transform
```

This functionality doesn't extend to the rake import task. Will address in a separate PR.


miguelff:
Changes in this PR introduced failures in tests, please check https://travis-ci.org/elastic/elasticsearch-rails/builds

PR #116, opened 2014-05-11
Compatibility for rails 3.2

beatjoerg:
Rails to serialize model names to symbols instead of strings. Fixed in more recent versions of Rails.


karmi:
@beatjoerg I'm terribly sorry, I can see from git blame that I've already added the `to_s` conversion as part of 64111a2b... Thanks for the patch and sorry for my oversight when doing that commit.

PR #121, opened 2014-05-21
ActiveRecord adapter after_commit hook on update

aackerman:
`ActiveRecord` adapter `after_commit` hook on update should only update the document if the record has changed as is the convention with ActiveRecord instances themselves.

I would be happy to write a test if this is an acceptable change. It's not completely clear from the README how to setup the db for the test suite so I had failures when I ran them and didn't take the time to figure it out.


karmi:
@aackerman Somehow I forgot about this for a year, you should have pinged me :) I agree with the reasoning, do you think we can add some tests for that?

aackerman:
I'm going to go ahead and close this, it's been 3 years, if I really needed it the work would have gotten done 3 years ago.

PR #138, opened 2014-06-05
Support 1.0-style multi_fields in the mapping DSL

rsutphin:
This branch updates the `mapping` DSL to have nested `indexes` calls produce embedded fields (instead of properties) for the core types where this is appropriate.

This mapping syntax is allowed starting with [elasticsearch 1.0][es10mf] as a replacement for `multi_field`.

[es10mf]: http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/_multi_fields.html

`type: "multi_field"` will still work as before; this change should not break backwards compatibility.

A workaround before this patch (or its equivalent) is merged is to define the nested fields using a hash. E.g.:

```
indexes :foo, type: 'string', fields: { raw: { type: "string", analyzer: "keywords" } }
```

With this patch, the same mapping can be written as:

```
indexes :foo, type: 'string' do
  indexes :raw, analyzer: 'keywords'
end

### DIFF
--- a/elasticsearch-model/lib/elasticsearch/model/indexing.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/indexing.rb
@@ -36,6 +36,9 @@ module Elasticsearch
       class Mappings
         attr_accessor :options
 
+        # @private
+        TYPES_WITH_EMBEDDED_PROPERTIES = %w(object nested)
+
         def initialize(type, options={})
           raise ArgumentError, "`type` is missing" if type.nil?
 
@@ -49,7 +52,7 @@ module Elasticsearch
 
           if block_given?
             @mapping[name][:type] ||= 'object'
-            properties = @mapping[name][:type] == 'multi_field' ? :fields : :properties
+            properties = TYPES_WITH_EMBEDDED_PROPERTIES.include?(@mapping[name][:type]) ? :properties : :fields
 
             @mapping[name][properties] ||= {}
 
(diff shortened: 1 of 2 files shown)


karmi:
@rsutphin The week was a bit longer then expected :) Many thanks, merged!

(I've actually removed the dynamically defined tests, I'm a bit freaked out by those, and added more explicit DSL-based tests. Please ping me if you'd like to change something in that.)

phantomwhale:
Thanks for this - was butting up against just this issue today. Is this PR going into a tagged release anytime soon ? Will just grab the master build into my project for now...

inkstak:
Hi, this commit breaks my app specs.
Why adding `nested` to `TYPES_WITH_EMBEDDED_PROPERTIES` ?

I use nested properties like this :

    indexes :users, type: :nested do
      indexes :name_badge: 
    end

to expect the following mapping :

    "user": {
      "type": "nested",
      "properties": {
        "name": {
          "type": "string"
        }
      }
    }

So now, `properties` becomes `fields`


According to https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-nested-type.html and https://www.elastic.co/guide/en/elasticsearch/reference/current/_multi_fields.html :

> the field type multi_field has been removed. Instead, any of the core field types (excluding object and nested) now accept a fields parameter.

So, when recreating mapping, ES ignore the fields parameter, and the final mapping is now :

    "user": {
      "type": "nested"
    }

jogaco:
How is this config file supposed to be used with `Elasticsearch::Persistence::Model` ?
I need to define custom analyzers and cannot find the way to do it

PR #162, opened 2014-07-08
Update Sidekiq example in the README

xiaoronglv:


karmi:
Thinking about it, let's keep the example in the README minimal and without any `klass.constantize.find(record_id)` magic -- interested people should discover the example application anyway... Closing this if it's fine.

Issue #88, opened 2014-04-27
Why I always have to force create_index! ?


tenaz3:
searchable.rb

``` ruby
FORMAT = "...."
settings :analysis => {
        :analyzer => {
            :city_name => {
                'tokenizer' => 'standard',
                'filter' => ['lowercase', 'asciifolding'],
                'type' => 'custom'}
        }
    } do
      mapping dynamic: 'false' do
        indexes :id, type: 'long', index: :not_analyzed
        indexes :name, type: 'string', boost: 1, analyzer: 'snowball'
        indexes :city, type: 'string', boost: 10, analyzer: 'city_name'      
      end
    end
```

When I run in console

``` ruby
Model.import
```

In my  meta data at http://localhost:9200/_plugin/head always map all fields and does not follow the rules as eg. city_name for field city. So when i run the code below, it works..Why?

``` ruby
Model.__elasticsearch__.create_index! force: true
Model.import
Model.__elasticsearch__.refresh_index!
```

PR #182, opened 2014-07-28
update document should update with hash from as_indexed_json.

### LINKED ISSUE #178: Update does not work for custom properties.
Hi.

I have models that have custom index properties defined in the mapping. The values for these properties are computed when as_indexed_json is called. However, update_index call in Elasticsearch::Model::Indexing checks for changed attributes and only selects those values from as_indexed_json missing all the custom index properties.

see the :custom_prop in below class.

class M
  field :id
  field :name, type: String

  mappings do
    indexes :id,        index: :not_analyzed
    indexes :name,  analyzer: 'standard'
    indexes :custom_prop
  end

  def as_indexed_json(options={})
    _hash = as_json(only: [name])
    _hash['custom_prop'] = "some computed value"
  end
end

dwkoogt:
update document should update with hash from as_indexed_json which could also contain derived attributes.
resolves #178
this also should resolve issues #88 #166 and #195

### DIFF
--- a/elasticsearch-model/lib/elasticsearch/model/indexing.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/indexing.rb
@@ -393,8 +393,8 @@ module Elasticsearch
         def update_document(options={})
           if changed_attributes = self.instance_variable_get(:@__changed_attributes)
             attributes = if respond_to?(:as_indexed_json)
-              indexed_json = self.as_indexed_json
-              (self.attributes.keys - changed_attributes.keys).map(&:to_s).each{ |k| indexed_json.delete(k) }
+              indexed_json = self.as_indexed_json.with_indifferent_access
+              (self.attributes.keys.map(&:to_s) - changed_attributes.keys.map(&:to_s)).each{ |k| indexed_json.delete(k) }
               indexed_json
             else
               changed_attributes
--- a/elasticsearch-model/lib/elasticsearch/model/indexing.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/indexing.rb
@@ -393,8 +393,8 @@ module Elasticsearch
         def update_document(options={})
           if changed_attributes = self.instance_variable_get(:@__changed_attributes)
             attributes = if respond_to?(:as_indexed_json)
-              indexed_json = self.as_indexed_json.with_indifferent_access
-              (self.attributes.keys.map(&:to_s) - changed_attributes.keys.map(&:to_s)).each{ |k| indexed_json.delete(k) }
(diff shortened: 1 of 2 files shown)


afn:
Any update on this?

Meekohi:
Hate to just +1 an issue, but this is the same problem I'm having as well and a little lost. Are there any work-arounds in the meantime for "forcing" a derived field to get sent along with ES updates?

Meekohi:
Okay dug around in this more today and personally I like the @dwkoogt solution. In my opinion this is a pretty significant and sneaky bug that should be fixed.
1. It is exactly the same if you don't use `as_indexed_json`
2. It is exactly the same if you use `as_indexed_json` but only with real attributes (i.e. that can be tracked in `changed_attributes`)
3. If you use computed attributes or methods in `as_indexed_json` (i.e. anything that can not be tracked in `changed_attributes`) it conservatively sends those along during updates.

@karmi "It certainly cannot be merged as is... We can absolutely have talk about the problem there."

Could you let us know what needs to be improved here to have it merged? I'd be willing to spend some time on this.

karmi:
Hi @dwkoogt, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-rails/pull/182.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

dwkoogt:
Hi @karmi, just updated my profile with the email address. thx.

elasticsearch-model/lib/elasticsearch/model/indexing.rb
```diff
@@ -393,7 +393,9 @@ def delete_document(options={})
         def update_document(options={})
           if changed_attributes = self.instance_variable_get(:@__changed_attributes)
             attributes = if respond_to?(:as_indexed_json)
-              self.as_indexed_json.select { |k,v| changed_attributes.keys.map(&:to_s).include? k.to_s }
```
Meekohi: Is this change for performance? Original version seemed more readable.

dwkoogt: My original change included with_indifferent_access method which is a rails extension which I removed here.

dwkoogt:
Btw, I rebased on master and Travis is failing on this: $ curl -sS https://snapshots.elastic.co/downloads/elasticsearch/elasticsearch-6.0.0-alpha1-SNAPSHOT.tar.gz | tar xz -C /tmp

Meekohi:
@karmi any updates on what needs to happen here?

PR #196, opened 2014-08-11
Call import on proxy object

bdmac:
elasticsearch-model is very careful to not override any method names on
model classes if they already exist. When this happens you have to use
the __elasticsearch__ proxy object directly. That's great but the
elasticsearch-rails import rake task does not respect that.

In our case there is a conflict with the activerecord-import gem which
defines an import method on ActiveRecord::Base. The way the rake task
was written it was ending up calling activerecord-import's version of
import instead of elasticsearch-model's version.

This change simply forces the rake task to use the proxy object.

### DIFF
--- a/elasticsearch-rails/lib/elasticsearch/rails/tasks/import.rb
+++ b/elasticsearch-rails/lib/elasticsearch/rails/tasks/import.rb
@@ -59,7 +59,8 @@ namespace :elasticsearch do
         rescue NoMethodError; end
       end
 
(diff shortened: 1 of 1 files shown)


karmi:
@bdmac I'm terribly sorry, but apparently I have already fixed it in 4e1a9f3 :/ I must have forgot about this PR at that time. Sorry again, I probably didn't search Github issues properly when doing that...

PR #207, opened 2014-08-14
remove duplicate article.rb-related content in expert template - caused ...

binaryphile:
...template install to fail

There was some repeated text in the expert template which was causing the install to fail.

It was attempting to add text to the article.rb file before that file existed.  I excised the offending code.

### DIFF
--- a/elasticsearch-rails/lib/rails/templates/03-expert.rb
+++ b/elasticsearch-rails/lib/rails/templates/03-expert.rb
@@ -111,17 +111,6 @@ gsub_file "app/models/authorship.rb", %r{belongs_to :article$}, <<-CODE
 belongs_to :article, touch: true
 CODE
 
-insert_into_file "app/models/article.rb", after: "ActiveRecord::Base" do
-  <<-CODE
-
-  has_and_belongs_to_many :categories, after_add:    [ lambda { |a,c| Indexer.perform_async(:update,  a.class.to_s, a.id) } ],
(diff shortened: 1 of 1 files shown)


epologee:
*bump*.

The 03-expert.rb template still doesn't work, and in the mean time it seems more template rows have been added (like a `gsub` on a non-existing `article_tests.rb` file) that cause the template to still fail.

dabit:
Are you running the templates in order?

The 02, 03... Templates won't run by themselves.

You have to run 01, then 02 on top of that, then 03 on top of that and so on.

karmi:
@binaryphile @epologee Yeah, @dabit is right -- you have to run the templates on top of each other. Are you sure you run them like that?

I think there's maybe an issue already open to put some kind of guards there, which would enforce it, and fail loudly. I must look into that.

epologee:
I was indeed not running the templates progressively. Haven't had the time to check, but I assume all will work as intended once I do so.

karmi:
@epologee Thanks! I'll put in some checks to prevent running the templates out of order, keeping this one open until it happens.

tylerjl:
The templates [now have checks to help steer people towards running these in order](https://github.com/elastic/elasticsearch-rails/blob/26563ee2b821e953fd48ab97e05f51da479b81dd/elasticsearch-rails/lib/rails/templates/02-pretty.rb#L3), so per @karmi's last comment I think we're good to close this now. 👍

PR #212, opened 2014-08-16
@ids is not an instance variable set anywhere

tpitale:
But, there is a method called `ids` which collections the ids from hits: https://github.com/elasticsearch/elasticsearch-rails/blob/104f91f24ba8d3dc7629b0eaf058329dc2cdc08c/elasticsearch-model/lib/elasticsearch/model/response/records.rb#L33

### DIFF
--- a/elasticsearch-model/examples/datamapper_article.rb
+++ b/elasticsearch-model/examples/datamapper_article.rb
@@ -50,7 +50,7 @@ module DataMapperAdapter
   #
   module Records
     def records
-      klass.all(id: ids) # in ruby 2.1.2 you have to make these into integers
+      klass.all(id: ids)
     end
 
     # ...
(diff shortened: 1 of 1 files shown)


karmi:
Hey @tpitale, blast from the past! :) You're right, the `@ids` maybe worked at certain point in DM, but it doesn't anymore. Can you please [sign the CLA](https://github.com/elastic/elasticsearch-rails/blob/master/CONTRIBUTING.md) so I can merge the patch?

(I understand this is a tiny change, but it's way easier for us to ask for the signature even for these... Thanks for understanding!)

tpitale:
I'll sign as soon as I'm back in the States. Doesn't seem to work on mobile right now (just says preview not available). I should be back the end of next week.

karmi:
@tpitale Cool! Take your time and thanks! (Yeah, I think the service works correctly only on a computer...)

tpitale:
Signed!

tpitale:
I just re-signed the CLA. But the checks don't seem to update.

PR #220, opened 2014-08-27
Add Callbacks to the example datamapper adapter

tpitale:
I added these callbacks in my code, and thought it might be useful for other people.


karmi:
Yeah, I agree, this is just a couple of lines, why not add them. Same as with the other PR, would you be so kind and sign the CLA please?

tpitale:
Signed!

PR #287, opened 2014-11-13
Make "Active Record Pattern" quickly accessible

ebouchut:


karmi:
This makes sense! Thanks for that, merged.

Issue #54, opened 2014-03-23
Import Data Faster


bcackerman:
Is there a way to have this command run faster `bundle exec rake environment elasticsearch:import:model CLASS='model' FORCE=y`by including any table joins?  My model has mapping object on it that includes another table currently.

netwire88:
Hi @karmi, has any user been able to replicate your bulk import example with Mongoid instead of ActiveRecord? I search all over the net but since Mongoid doesn't have `find_in_batches`, I'm not sure how to implement your example above.

karmi:
@netwire88 It should work exactly the same for Mongoid -- there's an adapter for Mongoid shipped.

karmi:
@netwire88 I mean, if you wanna write a custom importing code like that, just use the appropriate Mongoid methods, see eg. this comment, https://github.com/mongoid/mongoid/issues/1334#issuecomment-2419621

netwire88:
Thanks @karmi, does the default `import` method that comes with elasticsearch rails use bulk import?

PR #288, opened 2014-11-16
adds to documentation around importing

meesterdude:
I ran into some slow indexing with my app and found the vanilla way of importing to be too slow, but didn't see any mention of alternatives in the documentation. I later see #54 as a possible solution, only after reaching a similar conclusion through my own muddling about. I figured it made sense for the documentation to expand upon this, and save future users the same trouble. I pondered if maybe there was a better way to import out of the box too -- or maybe improve the `import` method; but this seemed like a simpler step.

### DIFF
--- a/elasticsearch-model/README.md
+++ b/elasticsearch-model/README.md
@@ -149,7 +149,40 @@ Article.import
 It's possible to import only records from a specific `scope` or `query`, transform the batch with the `transform`
 and `preprocess` options, or re-create the index by deleting it and creating it with correct mapping with the `force` option -- look for examples in the method documentation.
 
-No errors were reported during importing, so... let's search the index!
+
+#### Speeding up import
+
+If you have a lot of records with complicated indexing, you may find that importing takes quite a while. `#import` is great for starting out but is not optimized. There are generally two ways to speed things up:
+
+1. Use batching and threads/forks to perform multiple imports at once
+2. Bulk import and preload associations
+
+An example of #2 is as follows:
+
+```ruby
+    def self.bulk_import
+      self.preload(:tags,:files).find_in_batches do |indexables|
+        bulk_index(indexables)
+      end
+    end
+    
+    def self.bulk_index(indexables)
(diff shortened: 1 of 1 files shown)


karmi:
Hi Russel, apologies for the ridiculous delay. As much as I like long READMEs, I wouldn't like it to blow up beyond reason, and I think this doesn't fit with it.

I see two options:

1. Creating a Wiki page at Github with something like "Tips and Tricks for Importing"
2. Moving the content there
3. When you say "use batching and threads/forks", it's music to my ears, but I'd like to see a _working_ example for that -- otherwise it's just handwaving on our part, leaving the user with the implementation as homework. What do you think?

meesterdude:
@karmi no trouble. I have no trouble with a wiki page for it, but I don't think the wiki is enabled? I don't see it in the sidebar. 

I like factory_girls https://github.com/thoughtbot/factory_girl/blob/master/GETTING_STARTED.md over what devise has https://github.com/plataformatec/devise/wiki

karmi:
Hi, true -- I've enabled Wiki just now. In this case, I think we're shooting for a separate page, with specific, detailed information about importing, so what we need is that page + a Home page with only the link?

meesterdude:
I was thinking of creating an "all things indexing" so this can go there, along with a tip about refreshing the index which i think is missing from the docs too; i found it in a github issue. 

So, in that light there's a page for each major topic, and everything can fall under those in some sense. versus many wiki pages of varying relevance and quality. That would be my approach to it, what do you think? 

otherwise, yes I was going to give code examples and explain how to optimize and improve.

karmi:
Hi, undecided, I think that "many wiki pages on specific topics" make usually more sense than trying to come up with a "big chapter" upfront... I'd start with that specific topic on importing, and see where it leads..

karmi:
@meesterdude Closing this, if it's fine, let's work on that Wiki page? Can you please open an issue to communicate when you have something ready?

PR #289, opened 2014-11-17
Validate .update method

kkirsche:
Follow rails convention by returning false unless the item is valid.

### DIFF
--- a/elasticsearch-persistence/test/unit/model_store_test.rb
+++ b/elasticsearch-persistence/test/unit/model_store_test.rb
@@ -311,6 +311,43 @@ class Elasticsearch::Persistence::ModelStoreTest < Test::Unit::TestCase
         assert subject.update( {}, { script: 'EXEC' } )
       end
 
+      should "not update an invalid model" do
+        subject.expects(:persisted?).returns(true)
+        subject.expects(:id).returns('abc123').at_least_once
+
+        @gateway
+          .expects(:update)
+          .never
+
+        subject.instance_eval do
+          def valid?; false; end;
+        end
+
+        assert ! subject.update
+        assert ! subject.persisted?
+      end
+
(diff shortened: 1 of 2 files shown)


karmi:
This makes sense! Apologies for the ridiculous delay, you should have pinged me!

If you still have time & energy for the patch, can you please update it:

1. There's been a `:validate` option introduced in 1fd6c0b087adb13bd1a94e7e50cf4dd856080e37, can we play with it?

2. Can you please add a unit test for this?

kkirsche:
Updated with `:validate` option and unit tests. Please let me know if you have any questions, comments or concerns.

elasticsearch-persistence/test/unit/model_store_test.rb
```diff
@@ -311,6 +311,43 @@ def valid?; false; end;
...
+        @gateway
+          .expects(:update)
+          .never
+
+        subject.instance_eval do
+          def valid?; false; end;
+        end
+
+        assert ! subject.update
+        assert ! subject.persisted?
```
karmi: Have you run the test? `subject.expects(:persisted?).returns(true)` ... :)

karmi:
Fixed the tests in b660d6a2398d3db7544297a9f17609122bef8c26 and merged.

elasticsearch-persistence/test/unit/model_store_test.rb
kkirsche: Test suite of this project has never worked for me sadly so I haven't been able to outside of Travis CI

dgmora:
Unless I'm miss understanding something, this is not fixing the important cases. 

If you have stored a valid model, and then update with an invalid attribute, `valid?` will be true, as the model has yet not been changed, and there won't be validation

PR #295, opened 2014-11-25
[MODEL] Added, scope and where to limit imports for mongoid

### LINKED ISSUE #267: Mongoid: use scope or where to limit imports?
Hi,
Is there a way to use `scope` or `where` clause for Mongoid imports? I see it was made available for ActiveRecord via this update https://github.com/nono/elasticsearch-rails/commit/4a44bc750d22fa4972d2161e75fa43a200c5b818, but I wasn't able to do so via Mongoid, none of the following code works:

```
MongoidArticle.import(query: -> { where('status = active') })

MongoidArticle.where(status: 'active').import
```

hallelujah:
Closes #267 

As stated in the issue above, it will add the same options as ActiveRecord for Mongoid on import

### DIFF
--- a/elasticsearch-model/gemfiles/3.0.gemfile
+++ b/elasticsearch-model/gemfiles/3.0.gemfile
@@ -9,4 +9,4 @@ gemspec path: '../'
 
 gem 'activemodel',  '>= 3.0'
 gem 'activerecord', '~> 3.2'
-gem 'mongoid',      '>= 3.0'
+gem 'mongoid',      '~> 3.0'
--- a/elasticsearch-model/lib/elasticsearch/model/adapters/mongoid.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/adapters/mongoid.rb
@@ -69,10 +69,11 @@ module Elasticsearch
             query = options.delete(:query)
             named_scope = options.delete(:scope)
             preprocess = options.delete(:preprocess)
+            scope = scoped
 
-            scope = scoped.batch_size(options[:batch_size])
             scope = scope.__send__(named_scope) if named_scope
             scope = scope.instance_exec(&query) if query
+            scope = scope.batch_size(options[:batch_size])
 
             scope.each do |item|
               items << item
--- a/elasticsearch-model/lib/elasticsearch/model/adapters/mongoid.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/adapters/mongoid.rb
@@ -66,17 +66,25 @@ module Elasticsearch
             options[:batch_size] ||= 1_000
             items = []
 
-            all.each do |item|
+            query = options.delete(:query)
+            named_scope = options.delete(:scope)
+            preprocess = options.delete(:preprocess)
+
+            scope = scoped.batch_size(options[:batch_size])
+            scope = scope.__send__(named_scope) if named_scope
+            scope = scope.instance_exec(&query) if query
+
+            scope.each do |item|
               items << item
 
               if items.length % options[:batch_size] == 0
-                yield items
+                yield (preprocess ? self.__send__(preprocess, items) : items)
                 items = []
               end
             end
 
             unless items.empty?
-              yield items
+              yield (preprocess ? self.__send__(preprocess, items) : items)
             end
           end
 
--- a/elasticsearch-model/test/integration/mongoid_import_test.rb
+++ b/elasticsearch-model/test/integration/mongoid_import_test.rb
@@ -1,0 +1,155 @@
+require 'test_helper'
+
+begin
+  require 'mongoid'
+  session = Moped::Connection.new("localhost", 27017, 0.5)
+  session.connect
+  ENV["MONGODB_AVAILABLE"] = 'yes'
+rescue LoadError, Moped::Errors::ConnectionFailure => e
+  $stderr.puts "MongoDB not installed or running: #{e}"
+end
+
+if ENV["MONGODB_AVAILABLE"]
+  $stderr.puts "Mongoid #{Mongoid::VERSION}", '-'*80
+
+  logger = ::Logger.new($stderr)
+  logger.formatter = lambda { |s, d, p, m| " #{m.ansi(:faint, :cyan)}\n" }
+  logger.level = ::Logger::DEBUG
+
+  Mongoid.logger = logger unless ENV['QUIET']
+  Moped.logger   = logger unless ENV['QUIET']
+
+  Mongoid.connect_to 'mongoid_articles'
+
+  if defined?(BSON::ObjectId)
+    class BSON::ObjectId
+      def as_json(*args)
+        to_s
+      end
+
+      def to_json(*args)
+        as_json().to_json
+      end
+    end
+  end
+
+  if defined?(Moped::BSON::ObjectId)
+    class Moped::BSON::ObjectId
+      def as_json(*args)
+        to_s
+      end
+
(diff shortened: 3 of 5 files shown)


karmi:
Hi, sorry for the ridiculous delay... If you have the time & energy for this patch still, an you please reset the commits, and rebase against current master? I think there are couple of breaking changes here and there... I'll leave some comments on the diff as well.

elasticsearch-model/gemfiles/3.0.gemfile
```diff
@@ -9,4 +9,4 @@ gemspec path: '../'
 
 gem 'activemodel',  '>= 3.0'
 gem 'activerecord', '~> 3.2'
-gem 'mongoid',      '>= 3.0'
+gem 'mongoid',      '~> 3.0'
```
karmi: Hmm, an you please provide reasoning for the change, in commit message or via comment? 

I'd like to keep the gemfiles intact, since it frequently totally messes up everything and I have to spend hours debugging it.

elasticsearch-model/lib/elasticsearch/model/adapters/mongoid.rb
```diff
@@ -66,17 +66,26 @@ def __find_in_batches(options={}, &block)
             options[:batch_size] ||= 1_000
             items = []
 
-            all.each do |item|
+            query = options.delete(:query)
+            named_scope = options.delete(:scope)
+            preprocess = options.delete(:preprocess)
+            scope = scoped
```
karmi: What does `scoped` in Mongoid do? Wouldn't this respect any "default" scope? I don't wanna do that, it's best that we default to "all records" during import, and make people use the scope, query options to filter the records to import.

elasticsearch-model/test/integration/mongoid_import_test.rb
```diff
@@ -0,0 +1,155 @@
+require 'test_helper'
+
+begin
```
karmi: The Mongo test setup has been recently extracted to test_helper.rb, I think here it will break for you.

elasticsearch-model/test/integration/mongoid_import_test.rb
```diff
@@ -0,0 +1,155 @@
...
+  logger = ::Logger.new($stderr)
+  logger.formatter = lambda { |s, d, p, m| " #{m.ansi(:faint, :cyan)}\n" }
+  logger.level = ::Logger::DEBUG
+
+  Mongoid.logger = logger unless ENV['QUIET']
+  Moped.logger   = logger unless ENV['QUIET']
+
+  Mongoid.connect_to 'mongoid_articles'
+
+  if defined?(BSON::ObjectId)
```
karmi: Hmm -- why fiddle with `BSON::ObjectId` at all? This is not something people do in their app probably? Can we make it work in a different way?

elasticsearch-model/test/integration/mongoid_import_test.rb
```diff
@@ -0,0 +1,155 @@
...
+
+        context "Mongoid importing" do
+          setup do
+            class ::MongoidImportArticle
+              include Mongoid::Document
+              include Elasticsearch::Model
+
+              field :title,      type: String
+              field :views,      type: Integer
+              field :numeric,    type: String  # For the sake of invalid data sent to Elasticsearch
```
karmi: Why set it up as a `String`? We can always pass a string even with the `field` set? Or will Mongoid cast it to number?

elasticsearch-model/test/integration/mongoid_import_test.rb
```diff
@@ -0,0 +1,155 @@
...
+        as_json().to_json
+      end
+    end
+  end
+
+  module Elasticsearch
+    module Model
+      class MongoidImportIntegrationTest < Elasticsearch::Test::IntegrationTestCase
+
+        context "Mongoid importing" do
```
karmi: Great test otherwise, thanks!

elasticsearch-model/test/unit/adapter_mongoid_test.rb
```diff
@@ -75,13 +75,57 @@ def ids
...
 
     context "Importing" do
-      should "implement the __find_in_batches method" do
-        DummyClassForMongoid.expects(:all).returns([])
-
+      setup do
         DummyClassForMongoid.__send__ :extend, Elasticsearch::Model::Adapter::Mongoid::Importing
+      end
+
+      should "raise an exception when passing an invalid scope" do
```
karmi: Can you please move this after the "happy paths", ie. after line 128 in your patch?

karmi:
Hi @hallelujah, the discussion kinda died out -- let's return to it?

karmi:
Duplicated in #412, @hallelujah, I need you to get back to me and//or synchronize with @fabiob and merge your patches.

elasticsearch-model/gemfiles/3.0.gemfile
fabiob: I like this change, since Mongoid 4 will bring a lot of breaking changes. This PR is about bringing AR and Mongoid support closer together, so it makes sense that both are locked into their 3.x releases.

fabiob: Without this change, both gemfiles will test against Mongoid 4.x.

This change is needed so `3.0.gemfile` tests against Mongoid 3.x and AR 3.x, and `4.0.gemfile` tests agains Mongoid 4.x and AR 4.x.

elasticsearch-model/lib/elasticsearch/model/adapters/mongoid.rb
fabiob: Fixed in fabiob/elasticsearch-rails@1d9b63c58893eb29b1eb41589d60ca461136f13e

elasticsearch-model/test/integration/mongoid_import_test.rb
fabiob: Fixed in fabiob/elasticsearch-rails@0399f8eb19a72381fe1558d73274a93cb133936e

elasticsearch-model/test/integration/mongoid_import_test.rb
fabiob: I'd rather override `as_indexed_json`. Fixed in fabiob/elasticsearch-rails@4b949c308a2e465a7ba15ee2566493cba7425365

elasticsearch-model/test/integration/mongoid_import_test.rb
fabiob: Yes, Mongoid will cast the field to a number.

Switched to an untyped field, which is more realistic.

Fixed on fabiob/elasticsearch-rails@0bd8991c105b2c9d9961639dd50edeca3fc00865

elasticsearch-model/test/unit/adapter_mongoid_test.rb
fabiob: I don't get it.

fabiob:
Hi, I opened a new PR – #414 – with my changes and the fixed suggested by @karmi.

alup:
Is there any progress on this feature?

estolfo:
I believe [this commit](https://github.com/elastic/elasticsearch-rails/commit/eb476aae1ffe5e01db7561f52ceced68b4844869) addresses this pull request so I'm going to close this. Please comment if there's something missing.

PR #312, opened 2015-01-08
Added note about pagination callbacks to model README

zackfern:
I ran into a bit of an issue where it wasn't entirely clear how pagination was loaded. I had to check out one of the example Rails app which has a caveat:

```
# NOTE: Kaminari has to be loaded before Elasticsearch::Model so the callbacks are executed
```

I copied the note and slightly modified the phrasing so that hopefully nobody in the future to have to take the time to figure out why it isn't working.

Thanks!


karmi:
Thanks, Zack!, I slightly reworded the note :) and merged your patch.

PR #319, opened 2015-01-16
[RAILS] Fix error inserting Kaminari into Gemfile

### LINKED ISSUE #318: Rails app templates - Error when generating 02-pretty
I get the following error when generating 02-pretty:

````
    generate    kaminari:views
Could not find generator 'kaminari:views'. Maybe you meant 'migration' or 'controller' or 'css:assets'
Run `rails generate --help` for more options.
        gsub    app/views/kaminari/_paginator.html.erb
````

It appears the root cause is that the `kaminari` gem is never actually added to my `Gemfile`.

jpendry:
Fixes #318


karmi:
@jpendry Ah, thanks for fixing it!

Issue #276, opened 2014-10-29
Rails app templates - Error when generating expert use template


elmerwu:
I ran the basic and pretty templates in sequence with no issue, however the expert template shows me this error while compiling:

    Reading articles from gzipped YAML...
     (0.1ms)  begin transaction
    SQL (0.3ms)  INSERT INTO "articles" ("abstract", "content", "created_at", "published_on", "shares", "title", "updated_at", "url") VALUES (?, ?, ?, ?, ?, ?, ?, ?)  [["abstract", "What does the way you speak say about where you’re from? Answer the questions to see your personal dialect map."], ["content", ""]..."], ["created_at", "2014-10-29 20:04:42.496382"], ["published_on", "2013-12-20"], ["shares", 1], ["title", "How Y’all, Youse and You Guys Talk"], ["updated_at", "2014-10-29 20:04:42.496382"], ["url", "http://www.nytimes.com/interactive/2013/12/20/sunday-review/dialect-quiz-map.html"]]
     (2.3ms)  commit transaction
     (0.4ms)  rollback transaction
    rake aborted!
    ActiveRecord::StatementInvalid: SQLite3::SQLException: cannot rollback - no transaction is active: rollback transaction

When it completes I only have one article in the database and if I add a new one it does not show up on the search.  I'm not sure why this is happening, any advice?

pprzetacznik:
I've got exactly the same problem.

pprzetacznik:
I found solution for this one. You've got to have redis turned on before running ```rake db:seed```.

elmerwu:
Thanks @pprzetacznik , but after I run rake db:seed, I still only see one article result.  I checked the database and there are 1000s of articles and comments too... I ran rake db:seed a few times so there might be some duplicates, but it should still be more than one article on the index page.

karmi:
The `02-pretty` now fails with an error:

```
generate    kaminari:views
Could not find generator kaminari:views.
        gsub    app/views/kaminari/_paginator.html.erb
The template [https://raw.github.com/elasticsearch/elasticsearch-rails/master/elasticsearch-rails/lib/rails/templates/02-pretty.rb] could not be loaded. Error: No such file or directory @ rb_sysopen - /Users/karmi/Desktop/searchapp/app/views/kaminari/_paginator.html.erb
```

pprzetacznik:
[...]

PR #337, opened 2015-02-27
Update Regex matching Elasticseach gem so Kaminari will insert correctly

louim:
This should fix the issue with the 02-pretty template described in #276. I'm not sure that the main issue described in the issue is related.


karmi:
Thanks, @louim!, this is the correct solution, I got distracted and didn't get to the fix.

PR #339, opened 2015-03-03
Save pre-saved changed_attributes.

treby:
Hi!

I use `update_document` method in `after_commit` for transaction. In a controller, if we call `save` method twice or more times in `transaction` block, `before_save` overwrites `@__changed_attributes` variable. So elasticsearch does not reflect entire changes.

So I tried to keep the variable before overwrite.


karmi:
Hi, thanks for the catch! Can we add a test for this? I think an integration test is fine in this case.

treby:
Thank you for your response! I will write test for this change.

treby:
@karmi I wrote a test. Please review. Tell me anything if you have any problems or advices!

karmi:
Thanks, @treby, merged!

treby:
Thanks!

PR #342, opened 2015-03-05
[STORE] allowed the updated_at timestamp to persist to the database upon...

kaptron:
Because the `updated_at` timestamping happens after the save, `self[:updated_at]` is only applying an `updated_at` timestamp to the model in memory, and not actually persisting that timestamp to the database. So it seems that the order of these lines needs to be flipped, with the `gateway.save` happening second, which will actually persist the `updated_at` change to the database. This would be more consistent with the `update` and `touch` methods which do actually persist the change.

### DIFF
--- a/elasticsearch-persistence/lib/elasticsearch/persistence/model/store.rb
+++ b/elasticsearch-persistence/lib/elasticsearch/persistence/model/store.rb
@@ -53,9 +53,9 @@ module Elasticsearch
               options.update index: self._index if self._index
               options.update type:  self._type  if self._type
 
-              response = self.class.gateway.save(self, options)
-
               self[:updated_at] = Time.now.utc
+
+              response = self.class.gateway.save(self, options)
 
               @_id       = response['_id']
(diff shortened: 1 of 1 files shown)


karmi:
Ah, thanks for the catch! Can you add a test for this? I think just adding another integration test is fine in this case.

karmi:
Hi @kaptron, just pinging if you had time to add the tests? No time pressure, just curious :)

kaptron:
@karmi Have not yet -- but I should! Will try to get on top of that.

karmi:
@kaptron Take your time, just ping me when you have something please!

ryansch:
I took a crack at this in #389

karmi:
@kaptron I think I'll go with @ryansch's patch in #389, if that's OK.

kaptron:
Yep, sorry again for neglecting -- @ryansch's patch looks great!

karmi:
Thanks guys, merged the patch!

Issue #333, opened 2015-02-18
SystemStackError: stack level too deep error while executing import on child class using Single Table Inheritance


rbaghele:
require 'mongoid'
    Mongoid.connect_to 'articles'

    class Article
      include Mongoid::Document

      field :id,    type: String
      field :title, type: String
      field :first_name, type: String
      field :last_name, type: String
      attr_accessible :id, :title, :published_at
      include Elasticsearch::Model

       settings index: { number_of_shards: 1 } do
        mappings dynamic: 'false' do
          indexes :title, analyzer: 'english', index_options: 'offsets'
          indexes :first_name, type: "string"
          indexes :display_name, type: "string", as: "full_name"
        end
      end

      def full_name
  	    "#{first_name} #{last_name}"
      end

      def as_indexed_json(options={})
        as_json(
    	    except: [:id, :_id],
    	    methods: [:full_name]
    	    )
      end
    end

    class ChildArticle < Article
    end  
    ChildArticle.create id: '0', title: 'title for test 0', first_name: 'jhon', last_name: 'carter'
    ChildArticle.create id: '1', title: 'title for test 1', first_name: 'smith ', last_name: 'jones'

    
    ChildArticle.import    # giving error for child class import
    SystemStackError: stack level too deep
    from /home/user/.rvm/gems/ruby-2.1.5/gems/pry-0.10.1/lib/pry/pry_instance.rb:355

fonji:
I have the same problem here and found some stuff... I think it needs to be investigated, support and documentation for child classes with single table are unclear or lacking.

Anyway, if you don't want everything in the same index re-```include Elasticsearch::Model``` in the child classes (ChildArticle in your example) seems to resolve the problem.

Now, if you want to index everything in the parent class (Article here), and want to use ```Elasticsearch::Model::Callbacks```, it seems more complicated:
```Article.import``` will import ChildArticles into the index, but it seems that the callbacks won't maintain the index, so you might need to re-write those. Also, you might want to add the "method" ```_type``` in ```as_indexed_json``` to use it as a filter.

karmi:
Hey guys, I'm starting a two week vacation today, so I won't be able to investigate this properly. I agree that STI support is lacking, on the other hand, I don't personally believe in STI too much, so we must find some balance :)

With regards to:

> Now, if you want to index everything in the parent class (Article here), and want to use Elasticsearch::Model::Callbacks (...)

[...]

Issue #144, opened 2014-06-13
Model.import causes stack level too deep error


tomciopp:
Currently when I try to use the import class method provided by elasticsearch-model, I run into `SystemStackError: stack level too deep`. It looks like it runs into an error in irb and pry when they try to run `eval` on the code. This causes errors to occur in the rake tasks as well. I am running version 3.2.16 of ActiveRecord.

StepanKuzmin:
I've the same issue as @arfl described

salmanzameer:
@karmi @arfl @StepanKuzmin any updates on this problem with STI ? 

Did any of you get any chance to resolve it. Also if you can share some gist for STI search would be great. I am have very hard time with it.

Previously, I was working with tire and it was working like a charm.

karmi:
I'll try to look into that later in the month, focused on something else right now, sorry!

jimboslic3d:
I'm also hitting the same issue that @arfl called out. Look forward to seeing what you come up with @karmi . Thanks for the help!

jlaw90:
Hi guys,

I had this same issue in an abstract base class, by instead doing instance_eval on the subclass fixed it.

I added the following method:

```ruby
  def self.inherited(child)
    super
    
    child.instance_eval do
      # Elastic search functionality!
      include Elasticsearch::Model

      # Automatically update the index on change
      include Elasticsearch::Model::Callbacks
    end
  end
```

StepanKuzmin:
Thanks @jlaw90, your code solved this issue for me.

karmi:
Thanks for the code eample, @jlaw90! Do you have an idea how we could address this in the gem, either in code or the documentation?

There are too many mishaps happening with STI, see eg. #333, I need to solve it soon.

jlaw90:
@karmi I haven't looked at the source for this gem, but I guess the issue must be that when the module is included it's using the class name to then cache the index, type and other parameters.
The fix would be to change any of these cached variables to be dynamically calculated when they're used.

This is all conjecture though, I don't know if there's anything else funky going on

karmi:
@jlaw90 Thanks!, I'll have a look at that when I'm back onlilne properly.

karmi:
The underlying problem is with the proxied methods. The "practical" solution seems to be include the `Elasticsearch::Model` module in the child class as well, as outlined in https://github.com/elastic/elasticsearch-rails/issues/333#issuecomment-103940296

PR #344, opened 2015-03-06
Use parent's index_name and document_type from child models

yuku-t:
Suppose there are three models:

``` rb
class Company < ActiveRecord::Base
  include Elasticsearch::Model
end
class TechCompany < Company; end
class TradingCompany < Company; end
```

then previously they respond to "document_type" as follows:

``` rb
Company.document_type        #=> 'company'
TechCompany.document_type    #=> 'tech_company'
TradingCompany.document_type #=> 'trading_company'
```

This commit changes them as:

``` rb
Company.document_type        #=> 'company'
TechCompany.document_type    #=> 'company'
TradingCompany.document_type #=> 'company'
```

So, we can use search scope like this:

``` rb
Company.search('*')  # Returns both TechCompany and TradingCompany

# Make `TechCompany.search` automatically add a filter such as
# `{ must: { term: { type: 'TechCompany' } } }`
TechCompany.search('*')  # Returns TechCompany only
```

### DIFF
--- a/elasticsearch-model/lib/elasticsearch/model/naming.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/naming.rb
@@ -34,7 +34,7 @@ module Elasticsearch
           if @index_name.respond_to?(:call)
             @index_name.call
           else
-            @index_name || self.model_name.collection.gsub(/\//, '-')
+            @index_name || self.base_model_name.collection.gsub(/\//, '-')
           end
         end
 
@@ -58,7 +58,7 @@ module Elasticsearch
         #     Article.document_type "my-article"
         #
         def document_type name=nil
-          @document_type = name || @document_type || self.model_name.element
+          @document_type = name || @document_type || self.base_model_name.element
         end
 
 
@@ -68,6 +68,13 @@ module Elasticsearch
         #
         def document_type=(name)
           @document_type = name
+        end
+
+        private
+
+        # Get the base class's model_name
+        def base_model_name
+          respond_to?(:base_class) ? base_class.model_name : model_name
         end
       end
 


yuku-t:
Sorry, I've just agreed to CLA.

karmi:
Interesting idea! Couple of notes, I'm travelling these days: we definitely need unit tests for this, and this is probably related to all the "single-table inheritance" (STI) issues being reported for Rails integration. Can you look at them? It would be great to have integration tests covering these...

karmi:
Hello, I'll need some time to process the patch -- there are couple of things which stick out, for instance, we deliberately don't add `active_record` as a dependency to the main .gemspec, but keep it in separate Gemfiles, etc.

yuku-t:
Hi, sorry for my late reply.

I miss understood the code base - I thought elasticsearch-model is used by only elasticsearch-rails. elasticsearch-persistence also uses it!

> we deliberately don't add `active_record` as a dependency to the main .gemspec, but keep it in separate Gemfiles

As I noted above, I added the dependency due to my wrong understanding. I've just removed it and updated the patch.

I want to write tests for this as soon as possible. (I'm busy now to write an article for Japanese famous tech magazine about "How to use elasticsearch in Ruby". It refers both this repository and elasticsearch-ruby :smirk: )

karmi:
@yuku-t Great news about the article! :) Take your time, there's no rush. Just ping me please when you have something ready, and ping me again if I get sidetracked and fail to respond.

BTW get in touch with @johtani for community support in Japan!

karmi:
Hi @yuku-t, did you had some time to return to this? Can you also please review #144 and #333?

yuku-t:
Hi, I finished the work last weekend. I check the failing tests and referred issues.

Tonkpils:
I'm assuming these two are related? #332

Tonkpils:
So, I tested out this changeset and the issue with this approach is that it ignores the `index_name` and `document_type` of the `base_class` if it's set. It will always use the model name instead of the proper base_class index/document.

PR #129, opened 2014-05-26
[MODEL] Searching multiple models

AaronRustad:
This takes the example code that @karmi presented in https://github.com/elasticsearch/elasticsearch-rails/issues/10 and exposes it via `Elasticseach::Model#search` 

Basic usage is:
```ruby
response = Elasticsearch::Model.search('jaguar', [Animal, Automobile])
```

Issue #50, opened 2014-03-19
Multi-type search


balexand:
Elasticsearch has support for searching against multiple types in a single query. But the `Response::Response`, `Response::Results`, and `Response::Records` classes only support searching from a single model class. It would be nice if these classes could support multiple model classes.

It looks like this could be implemented by modifying `Response::Records#records`. Instead of including this method directly from the adapter [like this](https://github.com/elasticsearch/elasticsearch-rails/blob/619e5383cd3c707a31223aeebf13691ca995cb91/elasticsearch-model/lib/elasticsearch/model/response/records.rb#L26), this could be implemented like:

``` ruby
# PSEUDOCODE
def records
  records = results.group_by(&:type).map do |type, results_for_type|
    # determine class and adapter for type ...
    # get records from adapter ...
  end.flatten

  # sort records so they are in the same order as results ...

  records
end
```

Unfortunately, it looks like this would involve significant, backward-incompatible changes to all of the classes listed above.

If this is a feature that you think is worthwhile, then please let me know and I may be able to work on this. Thanks.

Issue #30, opened 2014-03-04
Searching multiple indexes


melhotiby:
Hi @karmi 

We are converting our application that is currently using [Tire](https://github.com/karmi/retire) and we want to move it over to elasticsearch-rails. We have all the indexes all set but the problem is converting the search to multiple indexes. 

Here was our Tire implementation

``` ruby
@search_items = Tire.search(['first_index', 'second_index', 'third_index', 'fourth_index'], {size: 20, load: true}) do |search|
  search.query{|q| q.string params[:q], default_operator: "AND" } if params[:q].present?
  search.filter :or, { :not => { :exists => { :field => :published} } },
   { :term => { :published => true } }
end
```

Looking at your [Elasticsearch-model](https://github.com/elasticsearch/elasticsearch-rails/tree/master/elasticsearch-model) documentation it appears we can probably use this approach `Elasticsearch::Model.client.search(q: 'Something')` but I am not 100% sure that searches all the indexes. Even if it does it appears that the return value is not ruby objects so making use of the return is unclear to us.

Issue #10, opened 2014-02-06
Searching multiple indices/types (equivalent to "Tire.search")


viperdezigns:
Unfortunately, I've spent possibly too much time looking through this new gem looking for  a top level search method. Is there a way to search without using a specific model? (Like Tire.search used to do)

For the most part, I'm using the base client methods to do my searching, however, I think I'd like to handle most of the response parts through this gem. (Such as Kaminari)

Possibly, is there a way to search via Elasticsearch::Model.client.search(search_hash) and use this gem's result formatting?

I'm partially new to this lower level of ES (coming from Tire) so hopefully you know what I'm looking for. Any help is much appreciated!!

Other than having somewhat of a learning curve, LOVING these new gems for ES

daino3:
Hello,

Just a quick thought (and I might be missing something here). If you need this functionality, rather than monkey patching Elasticsearch::Model, since Karmi's implementation above inevitably hits elasticsearch-api#actions/search. Why not simply initialize a client w/ 
```
c = Elasticsearch::Client.new
c.search q: 'frisky kittens', body: {highlight: {fields: {...} } }
```
If no indexes are passed, it searches all indexes. Then you can create your own helper methods to convert them into AR objects, etc. Would eliminate the need the ES-model altogether and maintaining monkey-patched code?

rossdakin:
:+1: @daino3 

I just implemented multi-model search this way (shared index, different types).

I referenced the client through `Elasticsearch::Model` just for grins (assuming default values for logging, etc. are easier to keep consistent that way):

```Ruby
index = 'my-index'
body = {
  query: { ... },
  highlight: { ... },
  suggest: { ... }
}

@response = Elasticsearch::Model.client.search(index: index, body: body)
```

micahlmartin:
@rossdakin That works perfectly for me too. I don't need the direct models integration. I think this should be documented in a more prominent place.

karmi:
Hi all, thanks for the discussion and the examples! I'll be travelling in the coming days, but I'll keep the issue in mind and will return to it as soon as possible (mainly finishing the work in the `dsl` branch of elasticsearch-ruby)

webattitude:
Hi, what about single model, but multiple time-based indices ?

webattitude:
is wildcard index_name allowed ?

solaleye:
Late to this convo, but really informative!!! Thanks everyone. 👍

PR #345, opened 2015-03-14
[MODEL] Support for querying multiple models

miguelff:
This PR adds support for multiple models search, as discussed in:
- https://github.com/elastic/elasticsearch-rails/issues/10
- https://github.com/elastic/elasticsearch-rails/issues/30 
- https://github.com/elastic/elasticsearch-rails/issues/50

A previous attempt of implementing this was done in:
- https://github.com/elastic/elasticsearch-rails/pull/129

But although the PR is open, it was rejected mainly due to:
- Lack of documentation
- Missing unit and integration tests
- An inconsistent Search response API: Even though `#results` was part of the *multimodel-search* response, `#records` was not.

This PR tries to address the problems outlined above, and also implement a performant hit-to-record deserialization.

Thank you very much to @AaronRustad [for his PR](https://github.com/elastic/elasticsearch-rails/pull/129), which motivated this one.

### DIFF
--- a/elasticsearch-model/lib/elasticsearch/model/adapters/active_record.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/adapters/active_record.rb
@@ -7,7 +7,7 @@ module Elasticsearch
       module ActiveRecord
 
         Adapter.register self,
-                         lambda { |klass| !!defined?(::ActiveRecord::Base) && klass.ancestors.include?(::ActiveRecord::Base) }
+                         lambda { |klass| !!defined?(::ActiveRecord::Base) && klass.respond_to?(:ancestors) && klass.ancestors.include?(::ActiveRecord::Base) }
 
         module Records
           # Returns an `ActiveRecord::Relation` instance
--- a/elasticsearch-model/test/integration/multiple_models_test.rb
+++ b/elasticsearch-model/test/integration/multiple_models_test.rb
@@ -1,0 +1,156 @@
+require 'test_helper'
+require 'active_record'
+
+Mongo.setup!
+
+module Elasticsearch
+  module Model
+    class MultipleModelsIntegraton < Elasticsearch::Test::IntegrationTestCase
+      context "Multiple models" do
+        setup do
+          ActiveRecord::Schema.define(:version => 1) do
+            create_table :episodes do |t|
+              t.string :name
+              t.datetime :created_at, :default => 'NOW()'
+            end
+
+            create_table :series do |t|
+              t.string :name
+              t.datetime :created_at, :default => 'NOW()'
+            end
+          end
+
+          class ::Episode < ActiveRecord::Base
+            include Elasticsearch::Model
+            include Elasticsearch::Model::Callbacks
+
+            settings index: {number_of_shards: 1, number_of_replicas: 0} do
+              mapping do
+                indexes :name, type: 'string', analyzer: 'snowball'
+                indexes :created_at, type: 'date'
+              end
+            end
+          end
+
+          class ::Series < ActiveRecord::Base
+            include Elasticsearch::Model
+            include Elasticsearch::Model::Callbacks
+
+            settings index: {number_of_shards: 1, number_of_replicas: 0} do
+              mapping do
+                indexes :name, type: 'string', analyzer: 'snowball'
+                indexes :created_at, type: 'date'
+              end
+            end
+          end
+
+          [::Episode, ::Series].each do |model|
+            model.delete_all
+            model.__elasticsearch__.create_index! force: true
+            model.create name: "The #{model.name}"
+            model.create name: "A great #{model.name}"
+            model.create name: "The greatest #{model.name}"
+            model.__elasticsearch__.refresh_index!
+          end
+
+        end
+
+        should "find matching documents across multiple models" do
+          response = Elasticsearch::Model.search("greatest", [Series, Episode])
+
+          assert response.any?, "Response should not be empty: #{response.to_a.inspect}"
+
+          assert_equal 2, response.results.size
+          assert_equal 2, response.records.size
+
+          assert_instance_of Elasticsearch::Model::Response::Result, response.results.first
+          assert_instance_of Episode, response.records.first
+          assert_instance_of Series, response.records.last
+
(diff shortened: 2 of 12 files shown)


miguelff:
Description updated. Build fixed. I already signed the CLA too, however, it's not reflected here.

karmi:
Hello Miguel, thanks for the patch, it looks very promising!

I went over the code a bit, and I like the approach, naming and implementation. I'm starting a two week vacation now, however, so I'd like to get back to it later, and evaluate it in more detail. Is that OK for you? Can you ping me in two weeks if this gets buried under hundreds of other tasks and e-mails after I return? :)

I have one plea, actually -- your patch modifies a lot of source code locations, some are unrelated to the feature (like fixing the lograge test). Would it be possible for you to do `git reset master` (that's a soft reset, no files will be changed), and then commit logical chunks of the feature? Then you can just force push `git push --force` to the `multimodel-support` branch, and the pull request will be automatically updated.)

If this would be too complicated, no problem, I can always do it by myself.

miguelff:
Of course @karmi, I will undo rebasing and commit again the changes.

And in relation to the timings, no problem. I know you've been busy with the query DSL last week, so don't worry. I will comment on the issue in a couple of weeks if I don't have any news from you.

karmi:
@miguelff Perfect, thanks!

miguelff:
Done splitting PR into meaningful commits

juankiz:
This PR really helps me +1
Thanks for the hard work @miguelff

miguelff:
@karmi Hope your holidays went well :wink:

elasticsearch-model/lib/elasticsearch/model/adapters/active_record.rb
```diff
@@ -7,7 +7,7 @@ module Adapter
       module ActiveRecord
 
         Adapter.register self,
-                         lambda { |klass| !!defined?(::ActiveRecord::Base) && klass.ancestors.include?(::ActiveRecord::Base) }
+                         lambda { |klass| !!defined?(::ActiveRecord::Base) && klass.respond_to?(:ancestors) && klass.ancestors.include?(::ActiveRecord::Base) }
```
karmi: Why remove the check for `ancestors`?

miguelff: `Multimodel.new` is an instance of Multimodel, and not an instance of Class, hence it doesn't respond to `ancestors`. An alternative approach will be that multimodel would implement:

```ruby
def ancestors
   []
end
```

miguelff: Or let Multimodel be an instance of class, which I found it a bit tricky.

elasticsearch-model/test/integration/multiple_models_test.rb
```diff
@@ -0,0 +1,156 @@
...
+            model.__elasticsearch__.create_index! force: true
+            model.create name: "The #{model.name}"
+            model.create name: "A great #{model.name}"
+            model.create name: "The greatest #{model.name}"
+            model.__elasticsearch__.refresh_index!
+          end
+
+        end
+
+        should "find matching documents across multiple models" do
```
karmi: Getting a "stack level too deep" exception here:

```
    ERROR (0:00:00.545) test_: Multiple models should find matching documents across multiple models. 
          stack level too deep
        @ lib/elasticsearch/model/adapters/multiple.rb:112
```

miguelff: Let me check if it's because this latest change: 

https://github.com/miguelff/elasticsearch-rails/commit/9a74f4dec3579ca09e6384e3be5159f009ce081a#diff-bc189b68e631fd3efba47681fce1287eL109

miguelff: I have added a commit fixing the offending change. And checked that test now passes. I don't know why CI is not giving feedback, though.

karmi:
@miguelff Did some minor tweaks and b26b67a and merged, many thanks!

miguelff:
:+1: Thank you @karmi :rainbow:

AaronRustad:
So happy to see this in. Thanks @miguelff & @karmi for completing the feature!

Msadek02:
@karmi @miguelff  - I dont see anything specific in the documentation for how to filter on multiple models
So for example i want to do something like
```ruby
Elasticsearch::Model.search(params[:search], [Conversation, Message]).results.to_a.map(&:to_hash)
=> [{"_index"=>"conversations",
  "_type"=>"conversation",
  "_id"=>"20",
  "_score"=>0.6684941,
  "_source"=>{"id"=>20, "subject"=>"This is my subject", "created_at"=>"2016-03-30T18:55:55.000-04:00"}},
 {"_index"=>"conversations",
  "_type"=>"conversation",
  "_id"=>"21",
  "_score"=>0.6684941,
  "_source"=>{"id"=>21, "subject"=>"Another subject ", "created_at"=>"2016-03-30T18:57:01.000-04:00"}}]

# I want to do something like this
Elasticsearch::Model.search(params[:search], [Conversation, Message], {term: {id: "21"}}).results.to_a.map(&:to_hash)

# which should only return this object
[ {"_index"=>"conversations",
  "_type"=>"conversation",
  "_id"=>"21",
  "_score"=>0.6684941,
  "_source"=>{"id"=>21, "subject"=>"Another subject ", "created_at"=>"2016-03-30T18:57:01.000-04:00"}}]
```

Is that possible to do?

miguelff:
> Is that possible to do?

As per [this line](https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-model/lib/elasticsearch/model.rb#L180), options are being forwarded to the search request. Then, they will [merged](https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-model/lib/elasticsearch/model/searching.rb#L39-L42) with search definition provided (`params[:search]`  in your example) and then run against Elasticsearch. 

So, yes, it should be possible, although is not a behavior that has been properly tested.

railsdevmatt:
@miguelff - I have encountered this issue as well, can you give a quick snippet to show what you mean?

PR #360, opened 2015-03-30
Update badges

abookyun:
Add Travis-CI & Code Climate badges!


abookyun:
According to CLA status, I'll close this first and resubmitting it later. maybe move the badges under "Running the Test Suite".

PR #361, opened 2015-03-31
[MODEL] Add Sequel adapter

badosu:
I am sending this prematurely to discuss whether this is wanted upstream or not before I continue working on it.

Basically: I've searched extensively for Sequel integration with ElasticSearch and found none, so I developed a partial adapter for `Sequel::Model`.

The following points need to be discussed if this is to be wanted:

- It must define methods to work with callback hooks, see the [hook docs](http://sequel.jeremyevans.net/rdoc/files/doc/model_hooks_rdoc.html). I made the `Callbacks` module a Sequel plugin, but let me know if the correct place would be elsewhere, e.g. `ext/sequel`.

- It must call `plugin(:active_model)` at the class instance. Currently this is not performed automatically by this adapter, let me know if this is desired. See the [plugin docs](http://sequel.jeremyevans.net/rdoc-plugins/classes/Sequel/Plugins/ActiveModel.html)

As soon as I receive feedback that this could be pulled I'll work on the missing methods and tests.

### DIFF
--- a/elasticsearch-model/lib/elasticsearch/model/adapters/sequel.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/adapters/sequel.rb
@@ -1,0 +1,92 @@
+module Elasticsearch
+  module Model
+    module Adapter
+
+      # The default adapter for models which haven't one registered
+      #
+      module Sequel
+        Adapter.register self,
+                         lambda { |klass| !!defined?(::Sequel::Model) && klass.ancestors.include?(::Sequel::Model) }
+
+        # Module for implementing methods and logic related to fetching records from the database
+        #
+        module Records
+
+          # Return the collection of records fetched from the database
+          #
+          def records
+            klass.find(@ids)
+          end
+        end
+
+        # Module for implementing methods and logic related to hooking into model lifecycle
+        # (e.g. to perform automatic index updates)
+        module Callbacks
+
+          # Handle index updates (creating, updating or deleting documents)
+          # when the model changes, by hooking into the lifecycle
+          def self.included(base)
+            base.plugin(self)
+          end
+
+          module InstanceMethods
+            def after_create
+              super
+              __elasticsearch__.index_document
+            end
+
+            def after_update
+              super
+              __elasticsearch__.update_document
+            end
+
+            def after_destroy
+              super
+              __elasticsearch__.delete_document
+            end
+
+            def as_json(opts)
+              to_hash
+            end
+          end
+        end
+
+        # Module for efficiently fetching records from the database to import them into the index
+        #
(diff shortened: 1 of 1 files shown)


karmi:
Hi @badosu, sorry for the delay. It's absolutely OK to include the Sequel adapter with the gem.

Regarding your points:

1. Callbacks -- Does this concern the "automatic" callbacks? I think there's no need to implement them just for the sake of doing so; Sequel is probably used by advanced users who will like to implement their own indexing routines anyway

2. Calling `plugin(:active_model)` -- not familiar with Sequel best pratices, but I don't think there's any harm doing this automatically, if the ActiveModel features such as naming are needed (which I presume).

yo-mash:
I think you may need to required your module here as well: 
https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-model/lib/elasticsearch/model.rb#L17

karmi:
Hi @badosu, just checking in if I can help with something. No pressure at all, we can leave this open!

badosu:
Hi @karmi , I was using this adapter as something necessary to a feature I was implementing. However elasticsearch was not needed after the prototype.

So since I did not require an extensive feature set using the adapter I never got to the point where it could mature. However, if you have a guideline of what are the missing bits that are needed to implement to make this adaptar mainline, let me know that I can finish the job.

karmi:
No worries! I think that if anybody needs the adapter, they can always write their own implementation, similar to your own. It should be easy to find via search in issues as well. Thanks!

badosu:
@karmi nice, thanks for the feedback. If you need any help with this on the future let me know.

mrbrdo:
I updated the adapter a little bit: https://gist.github.com/mrbrdo/61e6b3628acf57d94b6d
It would be much better if there was official support for this...

karmi:
@mrbrdo I'm all for having it as part of the bundle adapters -- feel free to pick up the PR, build your commits on top of @badosu's ones and send it!

karmi:
Just a note, to pull the original branch from the pull request to your local repository and work on top of that, you can do eg.:

```bash
git fetch -v origin pull/361/head:issues/361 && git checkout issues/361
```

badosu:
@mrbrdo Some remarks:

- Line 11: Every class and module responds to ancestors, you can remove that check
- Line 22: Since elasticsearch requires you to use active_model, the original implementation should work, unless this is new functionality
- Line 70: I am not sure if active_support is required to use a minimal set with elasticsearch, I tried to avoid using methods like `stringify_keys`

Thanks for the patch! It should be straighforward to cherry-pick this commit and perform your own changes! I'd love to see this merged! :-)

Issue #270, opened 2014-10-18
mapping not honoring document_type


andrejj:
I set a document_type directive.
Then use Influencer13.create_index!
But the mapping does not contain my document type.
Instead it contains a type derived from class name.

```ruby
class Influencer13
  include Elasticsearch::Persistence::Model
  index_name "influencers"
  document_type 'influencer'
  mapping do
    puts @type
    @type = 'influencer'
  end
end
```

=> "influencer13"

tomeara:
@karmi I'd like to consider contributing to this project. Would this bug be something that you need help with, or are you already looking into it?

karmi:
@andrejj Should be fixed by @miguelff.

@tomeara Terribly sorry to miss your message! Please do ping me *again* if I don't respond in time...

PR #366, opened 2015-04-11
[PERSISTENCE] Forward document type to mappings

miguelff:
This PR tries to fix the bug described in https://github.com/elastic/elasticsearch-rails/issues/270

The problem was the following.
* Once a model extends `Persistence::Model` it sets up [two attributes]( https://github.com/elastic/elasticsearch-rails/blob/b5eb4f6180e51f744ceac32ede0d9636231e4118/elasticsearch-persistence/lib/elasticsearch/persistence/model.rb#L122-125). As a consequence, the `attribute` macro instantiates a `Elasticsearch::Model::Indexing::Mappings` in the model. This instance contains an attribute `@type` that defaults to the model name in lowercase.
* Any other definition of `document_type` will override the document_type in the `gateway` delegate, but the `mappings` instance originally created in the previous step remains untouched, and hence in an inconsistent state.

What I did here is: Instead of simply forward `document_type` to the gateway, also change the mappings' `document_type` accordingly, leaving it in a correct state.

An integration test was modified, in order to have a regression of the fix.

### DIFF
--- a/elasticsearch-model/lib/elasticsearch/model/indexing.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/indexing.rb
@@ -34,7 +34,7 @@ module Elasticsearch
       # Wraps the [index mappings](http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/mapping.html)
       #
       class Mappings
-        attr_accessor :options
+        attr_accessor :options, :type
 
         def initialize(type, options={})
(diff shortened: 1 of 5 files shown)


karmi:
Waiting for unit tests, otherwise good to merge!

miguelff:
@karmi Done.

karmi:
Thanks, Miguel!, merged.

PR #335, opened 2015-02-24
Force no_timeout in mongoid adapter

prathe:
I can't index without no_timeout. Indexation of my model takes 5 hours.

PR #370, opened 2015-04-17
[MODEL] Fix wrong retrieval of records corresponding to indexed documents when querying multiple models

### LINKED ISSUE #369: [MODEL] Wrong retrieval of records corresponding to indexed documents when querying multiple models.
*This has been shown in a production deployment of elasticsearch-model v0.17.0*

In [this commit](https://github.com/elastic/elasticsearch-rails/commit/b26b67a9b6719177ac6b0265aef580448048a360) some refactorings where made to [the adapter for multiple records](https://github.com/elastic/elasticsearch-rails/commit/b26b67a9b6719177ac6b0265aef580448048a360#diff-bc189b68e631fd3efba47681fce1287e). 

One of the refactorings was a reduction of the cyclomatic complexity (and renaming of `__type_records`):

changes

#### This change drove to the following

* A user searches across several models. Along the search execution the code reaches the point above.
* A certain `klass` (say for instance an ActiveRecord model) wont' match any of the cases on the `case/when` statement, entering in the default clause.
* Then  (line 60 on the right side) `find` will be invoked over `klass` which is an ActiveRecord, but unluckily ActiveRecord will respond to the message. 
* Due to eventual consistency, some of the records returned by ES, are no longer present in the DB.
* So in this case, `klass.find(ids)` will raise an exception as no all the ids where found in the DB. :boom: 

I'm preparing a fix for this.

miguelff:
This PR fixes #369 

* The first commit adds a fix for a regression introduced in #335, to depart from a green build
* The second commit includes:
      * A refactor over a `case/when` statement, as for objects. `case/when` only matches the branches in which `===` returns true. i.e. an object that is instance of a class. But if the object is a class itself, or a module like the adapters, it won't enter any of the `when` branches.
      * Taking back the call to `compact` over the collection of records, because as per eventual consistency, ES can return some results, which corresponding records no longer exist. As a consequence, calling records would return an array with positions including nil.
      * An integration test checking the correctness of the two points above.
* The third commit cleans up a bit the integration tests for multiple model.

### DIFF
--- a/elasticsearch-model/lib/elasticsearch/model/adapters/multiple.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/adapters/multiple.rb
@@ -18,9 +18,11 @@ module Elasticsearch
           def records
             records_by_type = __records_by_type
 
-            response.response["hits"]["hits"].map do |hit|
+            records = response.response["hits"]["hits"].map do |hit|
               records_by_type[ __type_for_hit(hit) ][ hit[:_id] ]
             end
+
+            records.compact
           end
 
           # Returns the collection of records grouped by class based on `_type`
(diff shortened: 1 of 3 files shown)


karmi:
@miguelff Thanks, merged!

PR #388, opened 2015-05-08
Only mention ActiveJob in the README

tilsammans:
Since ActiveJob is standard now, we can use that to document the background worker.

I moved the Sidekiq-specific example to the Wiki.

### DIFF
--- a/elasticsearch-model/README.md
+++ b/elasticsearch-model/README.md
@@ -471,59 +471,38 @@ end
 #### Asynchronous Callbacks
 
 Of course, you're still performing an HTTP request during your database transaction, which is not optimal
-for large-scale applications. A better option would be to process the index operations in background,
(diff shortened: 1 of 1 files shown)


karmi:
Hi, thanks, but I'll need to think about this a little bit. Sidekiq is just an example, should be straightforward to adapt for any system such as Resque, or ActiveJob. While I applaud abstract adapters in general (MultiJson, Faraday), not _so_ sure about ActiveJob...

karmi:
Hi, I'm sorry, but I've decided to keep the current README. I like the complex example, and I like Sidekiq as a model for asynchronous processing. Thank you for the time you take to make all the changes!, but I really want to keep what we have. Let's revisit it in the future!

tilsammans:
No worries.

PR #393, opened 2015-05-11
Stop requiring empty block for mapping options

ryansch:
This just bit me when trying to configure a class that includes Elasticsearch::Persistence::Model.  Since the attribute definitions handle setting up the mapping, I just wanted to set `dynamic` and `_routing`.  I had to resort to debugging to find out that I had to pass an empty block to get the `#mapping` method to actually do something with my options.


karmi:
@ryansch Argh!, this is so obvious!, thanks for catching and fixing it!

PR #395, opened 2015-05-13
[WIP] Convert from Virtus to ActiveAttr

ryansch:
I've had so many issues with the way Virtus decides to do things that I finally gave up and ripped it out in favor of ActiveAttr.

PR #407, opened 2015-05-22
Fix timestamp tests

ryansch:
Until we have a version of rails that uses millisecond precision (>=4.1)
we have to wait an entire second to avoid race conditions.

See https://github.com/elastic/elasticsearch-rails/pull/395/files#diff-beb7964b74f717683049a9437a5570a2R1 for a breakdown on the differences in timestamp precision in different versions of rails.


ryansch:
@karmi Fixed the failing tests on master.  Need to revert your sleep changes.

karmi:
Thanks, Ryan!, I guess I got carried away by the one green run when I tweaked your patch and wasn't thinking :)

ryansch:
:-)

Issue #382, opened 2015-04-29
Problems with pagination in MultiModel Search


miguelespada:
I try to use pagination with Kaminari in a MultiModel Search.
    
This is my call:
    Elasticsearch::Model.search(query, [AAA, BBB]).page(1).results

This is the error:
     Failure/Error: response = MultiModelSearch.search('xxx')
     NoMethodError:
       undefined method `default_per_page' for #<Elasticsearch::Model::Multimodel:0x007ff9dbe4f520>

Is pagination supported for multimodel?

xpepermint:
+1

miguelespada:
It works with will_paginate

Elasticsearch::Model.search(query, [AAA, BBB]).per_page(25).page(1).results

karmi:
@miguelespada Don't know from top of my head, we'll look into that.

@miguelff, would you spare some time to look at the issue, please?

miguelff:
Sure, I will look into as soon as I got some time.

miguelff:
One question @miguelespada, does the Gemfile of your application declare `gem 'kaminari' ` **before** elasticsearch-rails?

xpepermint:
Btw... I have `kaminari` before `elasticsearch-rails`.

miguelespada:
Hi @miguelff,
I do have kaminari before ES

# gem "will_paginate_mongoid"
gem "kaminari"
gem 'elasticsearch-model', git: 'git://github.com/elasticsearch/elasticsearch-rails.git'
gem 'elasticsearch-rails', git: 'git://github.com/elasticsearch/elasticsearch-rails.git'

kikyous:
+1

kikyous:
@miguelespada , @miguelff 
add `default_per_page` method simply works for me

```ruby
module Elasticsearch
  module Model
    class Multimodel
      def default_per_page
        10
      end
    end
  end
end
```

miguelff:
Thanks for the feedback, I will tackle the issue as soon as I got some time. Before the end of the week.

miguelff:
Working on this.

xpepermint:
thx!

miguelff:
Waiting for @karmi's review

karmi:
Should be solved by @miguelff patch in #408 now!

miguelespada:
Thanks! @miguelff  and @karmi

rkrdo:
Are you guys planning on releasing a version on rubygems with this patch?
[...]

PR #408, opened 2015-05-24
[MODEL] Proper defaults for pagination in multimodel searches

miguelff:
This PR implements a fix for #382 

Due to the fact that different models aggregated on a multimodel search, could have different default values for the page size. I have opted for defaulting to Kaminari's and WillPaginate's respective default values.

I have included tests for ensuring correct pagination using both toolkits.


karmi:
@miguelff Thanks!, added a simple integration test and merged!

miguelff:
:sparkles:  Thank you @karmi :sparkles:

Issue #421, opened 2015-06-10
[MODEL] Multiple model searches with Index Alias fails


AaronRustad:
I'm searching across multiple models in the common way:

```ruby
Elasticsearch::Model.search(my_query, [Content, Company]).records.first
```
When we deploy to staging where we use aliases we have problems. Elasticsearch will return a valid document, however, [Elasticsearch::Model::Adapter::Multiple::Records#__type_for_hit](https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-model/lib/elasticsearch/model/adapters/multiple.rb#L96) can't identify the appropriate model since `_index` returned from Elasticsearch is populated with the actual index name (`content-staging`), rather than the alias.

Is there a known strategy for getting around this? I'd hate to have to define actual index names in my models and lose all the benefits of aliases.

miguelff:
We have exactly the same issue in our current deployment. Index aliases are not bi-jective, in the sense that an alias can denote different indexes, so for the moment, this is a current limitation of the future.

In our particular scenario, we have workaround it by monkeypatching `Elasticsearch::Model::Adapter::Multiple::Records`:

```ruby
  #
  # This monkey-patch adds support to deserializing records that belong to an
  # aliased index.
  #
  # Inside the memoization, original implementation was:
  #
  # `Registy.all.detect { |model| model.index_name == hit[:_index] && model.document_type == hit[:_type] }`
  #
  # Which doesn't seem to be suitable for aliased indexes. In our case, we assume we are not going to
  # have two different models, named the same way in two different namespaces, so we can avoid the first
  # condition.
  #
  def __type_for_hit(hit)
    @@__types ||= {}
    @@__types[[hit[:_index], hit[:_type]].join("::")] ||= begin
      Elasticsearch::Model::Registry.all.detect { |model| model.document_type == hit[:_type] }
    end
  end
```

miguelff:
So basically, if you don't have different models with the same name, under different namespaces, it's enough information to deserialise the records, that coming from the document type. As by convention, the document type for a model, will be its name

AaronRustad:
That's an interesting solution. I was thinking of doing something like defining a pattern that your indexes follow. 

```ruby
Registy.all.detect { |model| 
  (model.index_name == hit[:_index] || model.index_pattern =~ hit[:_index]) 
    && model.document_type == hit[:_type] 
}
```

And in my model, it would look something like:

```ruby
  def self.index_pattern
[...]

PR #428, opened 2015-06-15
Add the ability to define how models are mapped to hits.

AaronRustad:
This PR addresses https://github.com/elastic/elasticsearch-rails/issues/421

When searching against aliases, results from `Elasticsearch` are returned with the actual `_index` name rather than the alias name. The current implementation of Multiple Model Search uses the `_index` to determine which model to use. Unfortunately, when working with an aliased index, this mapping fails.

I don't think there is a 'correct' way to get around this, other than to give the user of the library the option of setting the logic for determining how to map indices to aliased models.

This PR gives the option of overriding the default `model` to `hit` selection code.

For example, in production, I'll use a timestamped index like `content-20150515164511'. 

``` ruby
# INITIALIZER
Elasticsearch::Model.model_to_hit_selector = lambda do |model,hit|
  /#{model.index_name}-.*/ =~ hit[:_index] && model.document_type == hit[:_type]
end
```

### DIFF
--- a/elasticsearch-model/lib/elasticsearch/model.rb
+++ b/elasticsearch-model/lib/elasticsearch/model.rb
@@ -179,6 +179,30 @@ module Elasticsearch
         request = Searching::SearchRequest.new(models, query_or_payload, options)
         Response::Response.new(models, request)
       end
+
+      # Define a rule for mapping hits to models when searching across multiple models
+      #
+      # @param model_to_hit_selector [lambda(model,hit)] - Must return a boolean
+      #
+      # @example Map indices to models that operate against aliases
+      #
+      #     Elasticsearch::model_to_hit_selector = lambda do |model, hit|
+      #       /#{model.index_name}-.*/ =~ hit[:_index] && model.document_type == hit[:_type]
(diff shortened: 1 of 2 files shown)


AaronRustad:
Is there any interest in seeing this merged... I know I am!  ;-)

hongsheng-razoo:
I am facing the exact same problem ... can we merge this solution?

karmi:
Hi, ugh, this is one of the PRs which has slipped by... I was concentrated on other things for a long strecth of time, and in the last weeks was more focused on the Ruby client (preparing for the 5.0 release and similar).

I like the idea of a lambda, @AaronRustad, you know I do :) I'll look into this today or tomorrow, please bug me until we sort this out. Sorry to put the strain on you, but it would help me...

gareth:
This error has just bitten us too, I'm definitely interested in getting a solution merged

AaronRustad:
Lol, I haven't bugged you for a while @karmi .... I've since moved on and I'm not sure there is any interest in this any more. Shall we close it?

PR #448, opened 2015-08-01
Add __elasticsearch__ to model instance in docs

rusikf:
Article instance doesn't have direct method `as_indexed_json`, only through proxy `__elasticsearch__`


BenMorganIO:
:+1:

tylerjl:
Fixed in 0afd8b8

PR #451, opened 2015-08-04
Remove extra level from jbuilder code.

denispeplin:
The original jbuilder code produces:

``` json
{"query":{"match":{"title":{"query":"fox dogs"}}}}
```

And it actually should be:

``` json
{"query":{"match":{"title":"fox dogs"}}}
```


denispeplin:
I signed the agreement after making this pull request.

PR #464, opened 2015-09-03
[MODEL] Fixes searches with multiple models and aliases

dabit:
It is typical for production environments to use index aliases
to simplify maintenance.

For example, if you had an index for the `Article` model, your index
could be named `articles1` and then you have an alias `articles` that
points to that index.

This way if you need to recreate your index without downtime, you can
create and populate another index called `articles2` and simply change
the alias when you are done.

The current code in the multiple adapter asumes the index name declared
in the model will match the one returned by elasticsearch in the hits response
but this is not the case. When using an alias, elasticsearch returns the
actual index name, not the alias.

What I am doing here is match the index name in the model and the index name
in the hit by prefix. So now the match happens even if the name of the index
has more characters after the name.

Meaning, if the index name defined in the model is `articles` and the
hit returns `articles999999` it will be a match and everything will be
allright.

### DIFF
--- a/elasticsearch-model/lib/elasticsearch/model/adapters/multiple.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/adapters/multiple.rb
@@ -93,7 +93,7 @@ module Elasticsearch
 
             @@__types[ "#{hit[:_index]}::#{hit[:_type]}" ] ||= begin
               Registry.all.detect do |model|
-                model.index_name == hit[:_index] && model.document_type == hit[:_type]
+                hit[:_index] =~ /#{model.index_name}/ && model.document_type == hit[:_type]
               end
             end
           end
--- a/elasticsearch-model/test/unit/adapter_multiple_test.rb
+++ b/elasticsearch-model/test/unit/adapter_multiple_test.rb
@@ -102,5 +102,34 @@ class Elasticsearch::Model::MultipleTest < Test::Unit::TestCase
         assert_equal [2, 2, 1, 1, 3], records.map(&:id)
       end
     end
+
+    context "when using index aliases" do
+      setup do
+        @aliased_hits = [{
+            _index: 'dummy1',
+            _type: 'dummy_two',
+            _id: '3'
+          }, {
+            _index: 'other_index_99',
+            _type: 'dummy_two',
+            _id: '1',
+          }
+        ]
+        @multimodel.class.send :include, Elasticsearch::Model::Adapter::Multiple::Records
+        @multimodel.expects(:response).at_least_once.returns(stub(response: { 'hits' => { 'hits' => @aliased_hits} }))
+      end
+
+      should "match the right models" do
+        assert_instance_of Module, Elasticsearch::Model::Adapter::Multiple::Records
+        records = @multimodel.records
+
+        assert_equal 2, records.count
+
+        assert_kind_of ::Namespace::DummyTwo, records[0]
+        assert_kind_of ::DummyTwo,            records[1]
+
+        assert_equal [3, 1], records.map(&:id)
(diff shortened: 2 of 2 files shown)


miguelff:
Hey @dabit, thanks for your pull request, being able to do multi-model searches over aliased indexes is indeed something we need!

I have some concerns, though:
- On the one side, the implementation can crash when you are querying multiple models whose names overlap (ex. titles, subtitles)
- Unmarshalling is an operation done many times per query (as many as records being unmarshalled) however once we are able to check that a certain index_name, corresponds to a model, we can avoid doing regexep tests again and again.  Being able to cache that inference for all the queries will improve the performance of unmarshalling with little effort.

In my opinion we need to not depend exclusively on a convention here to do the matching. Because it won't work if aliasing is done following other criteria different from  the convention. As a suggestion, I would explore having some sort of information in the models about the aliases, that if present would be used to override the default.

What do you think, @dabit?

dabit:
Hi @miguelff 

I thought about the two concerns that you mentioned. 

Regarding the overlapping of index names, you are right. I can't think of a way to do this without risking bumping into this problem, except for being extra careful about it.

The problem with relying on the models for the alias name, for example setting something like `index_alias_name` that works similar to `index_name` is that, well, you could just use the alias directly as the index name and make the problem go away. But then, you'd need to restart your Rails apps when you change the index name to set the new one, so the zero downtime part would be gone.

Regarding the constant unmarshalling, perhaps I'm not understanding but I thought this line does some sort of caching to avoid matching the model with the index returned by the hit every time: https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-model/lib/elasticsearch/model/adapters/multiple.rb#L94

miguelff:
> Regarding the constant unmarshalling, perhaps I'm not understanding but I thought this line does some sort of caching to avoid matching the model with the index returned by the hit every time: https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-model/lib/elasticsearch/model/adapters/multiple.rb#L94

Gosh! @dabit, I didn't remember that (which is weird, because I wrote that code :see_no_evil:) you are right, sorry!



> The problem with relying on the models for the alias name, for example setting something like index_alias_name that works similar to index_name is that, well, you could just use the alias directly as the index name and make the problem go away. But then, you'd need to restart your Rails apps when you change the index name to set the new one, so the zero downtime part would be gone.

Another approach will be to be able to invalidate that (in the line of Rails.cache.clear) I think we would need to evaluate the tradeoff between flexibility and robustness (I actually have indexed titles and subtitles in an app!) and high-availability / zero-downtime.

What do you think, @karmi ?

jogaco:
I just hit the same problem and this needs to be addressed for a production environment.
What about a stricter regex?

```
hit[:_index] =~ /\A#{model.index_name}\d*\z/ && model.document_type == hit[:_type]
```
This reduces the risk of model name overlapping. It will only match exact index name ended by optional digits.

karmi:
I'm for adding some support for the requirement described by @dabit, but I'm also for a more stringent check/regex in this case. Something as simple as `/^#{model.index_name}[-_]?/` could be a start. This regex can also be exposed via a setting, something like `index_alias_pattern`, so users can customize it for their particular situation.
-----
Code examples:

```ruby
"articles".match /^articles[-_]?/
"articles-v2".match /^articles[-_]?/
"articles-2015-11-01".match /^articles[-_]?/
"articles_v2".match /^articles[-_]?/
```

jogaco:
This issue should also be addressed for elasticsearch-persistence
El dia 08/11/2015 10.42, "Karel Minarik" <notifications@github.com> va
escriure:

PR #466, opened 2015-09-04
use rails logger instead of calling stout directly

BenMorganIO:
This PR switches out the use of the `puts` method in several places in favour of the [Rails Logger](http://guides.rubyonrails.org/debugging_rails_applications.html).

I've noticed that when I boot my rails project up, there's tons of logs for elasticsearch. For production, the logs for successes don't need to be printed half of the time. This means that this PR also provides a.. wait for it... *performance improvement*!!! (How much? Well I have no idea, but you can research more here: [Impact of Logs on Performance](http://guides.rubyonrails.org/debugging_rails_applications.html#impact-of-logs-on-performance).)

This PR may cause failing specs since I went a bit haywire in switching out `puts` with a `Rails.logger.some_log_level`. Will revert those changes when I see them in the CI.

### DIFF
--- a/elasticsearch-model/README.md
+++ b/elasticsearch-model/README.md
@@ -248,7 +248,7 @@ When you want to access both the database `records` and search `results`, use th
 (or `map_with_hit`) iterator:
 
 ```ruby
-response.records.each_with_hit { |record, hit| puts "* #{record.title}: #{hit._score}" }
+response.records.each_with_hit { |record, hit| Rails.logger.info "* #{record.title}: #{hit._score}" }
 # * Quick brown fox: 0.02250402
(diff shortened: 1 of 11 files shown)


BenMorganIO:
Signed the contributor agreement.

karmi:
Hi, I'm all for using the `Rails.logger` where appropriate, but the most changes I see are in the documentation or examples, where's the use of `puts` deliberate (and I think warranted). That would also apply to the Rake task. The disputable place is in the `indexing.rb`, but both of those condition branches shouldn't appear often.

In other words, when you say:

> I've noticed that when I boot my rails project up, there's tons of logs for elasticsearch

Do you have an example?

PR #469, opened 2015-09-16
[MODEL] Added update_mapping! method. This method will update the map…

shikhar-j:
In a production environment, many models have their own mappings and they are likely to get updated once in a while. The create_index! method deletes all previous mappings which is counter productive and very risky for a production. Hence, I have added an update_mapping! method to Elasticsearch::Model. This way, one can simply update the mapping without damaging or erasing existing mappings.

I hope this is a valuable addition.

### DIFF
--- a/elasticsearch-model/lib/elasticsearch/model/indexing.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/indexing.rb
@@ -297,6 +297,28 @@ module Elasticsearch
             end
           end
         end
+
+        # Updates mappings with latest mappings
+        # Also creates the index if it doesn't exists already
+        # `mappings` defined in the model
+        #
+        # @example Update index for the `Article` model
+        #
+        #     Article.__elasticsearch__.update_index!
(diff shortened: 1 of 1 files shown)


shikhar-j:
I have already signed the CLA.

karmi:
Hi,

yes, true about the consequences of `create_index!`. However, I've deliberately stayed away from any "update mappings in ES based on mappings in model", because that opens up a huge avenue for misunderstanding on the user's part. For instance, you'd change the analyzer in the model, not realizing that this is not possible, calling some `update_mapping!` method, error is returned, you wonder what's going on...

In my experience, ES schema in a complex application must be decoupled from the models and the simple methods anyway, and created in some other components. This is particularly true when you have multiple models stored in the same index as different types (see the example application).

Also, the `client.indices.put_mapping ... body: self.mappings.to_hash` command is so simple, that there isn't much benefit in wrapping it in the gem code...

What do you think?

PR #470, opened 2015-09-21
Rails 4.1+ wants column selection on count

antifun:
Causes the annoying 

```
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
```

in the output/logs when attempting to import.


karmi:
Thanks! However, since the task should be framework agnostic (eg. importing Mongoid models), does it make sense to be a little bit granular here? Ie. checking if `klass` is ActiveRecord of certain version and such?

PR #483, opened 2015-10-25
[MODEL] Add "suggest" method to get response suggest

nipe0324:
I would like to use suggest of Elasticsearch response.
But `Elasticsearch::Model::Response::Response` class hasn't the method. That's why I added it.

I have already signed the CLA.

### DIFF
--- a/elasticsearch-model/lib/elasticsearch/model/response.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/response.rb
@@ -71,6 +71,12 @@ module Elasticsearch
         def aggregations
           response['aggregations'] ? Hashie::Mash.new(response['aggregations']) : nil
         end
+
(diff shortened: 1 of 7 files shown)


karmi:
I think it makes sense. Can we also have integration tests for the feature, and update to the [example application HTML template](https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-rails/lib/rails/templates/index.html.dsl.erb#L50) (in a separate commit?)

karmi:
Some tests on Travis are [failing](https://travis-ci.org/elastic/elasticsearch-rails/builds/87294328), do they run locally for you?

nipe0324:
I fixed it.

karmi:
I'm sorry for such a long delay, I've tweaked the code a bit and merged it! Thanks!

PR #488, opened 2015-11-05
Fix spelling.

will-in-wi:
Tiny fix in readme. Thanks for your work on this!


karmi:
Hi, thanks for catching that!, I always misspell it :) Do you think you can sign the [CLA](http://www.elasticsearch.org/contributor-agreement/)? I know this is only a tiny change, but it makes it much more easier procedurally for us to have it for all commits...

will-in-wi:
I've signed the CLA. Thanks!

tylerjl:
Fixed in 8f905c5

Issue #489, opened 2015-11-06
[MODEL] Decouple from order on which pagination libraries are required.


miguelff:
When using elasticsearch-model with a pagination library (i.e. Kaminari, or will-paginate) it's necessary to load those libraries before elasticsearch-model.

This leads to situations like [the one described in #487](https://github.com/elastic/elasticsearch-rails/issues/487#issue-115380728) in which the system would not behave as expected just because of not placing dependencies in the correct order.

This issue is aimed at discussing alternatives to remove the aforementioned coupling, and make elasticsearch-rails better at integrating with pagination libraries.

karmi:
@miguelff Nice try :), but I don't yet see a nice solution for this -- we rely on the fact that some pagination module is [defined](https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-model/lib/elasticsearch/model.rb#L38). Of course there could be some imperative way of including the pagination modules in the response, but I wonder how that would help -- the fact that you must load pagination libraries before the -model gem is quite clearly in the [README](https://github.com/elastic/elasticsearch-rails/tree/master/elasticsearch-model#pagination). Of course, we can make it **bold** :)

miguelff:
Couldn't we delay the inclusion of the pagination module? The thing is that we include the pagination at the precise point when we load elasticsearch-model. Wouldn't it make a difference if we include it later, maybe lazily?

karmi:
But what would be the event to hook into?

miguelff:
I don't know yet. For me this is the main concern of this issue: To find another point in time in which to include the pagination module. I will try to come up with something.

miguelff:
What about delaying the load of the libs until the actual `Response::Response` constant is resolved?

https://github.com/elastic/elasticsearch-rails/compare/master...miguelff:master

Is there any reason (a-priori) to discard something like this?

miguelff:
Sorry, I was wrong about this. I will do more tests and come back to you again.

karmi:
The tricky thing with this is that it can introduce delays in parts of code. I'd rather stay away from this, since what we're trying to solve here is solved when people read the documentation. It's no surprise that in Ruby, the order in which you include files/libraries matters. I was only half-joking when I suggested we make the instruction bold .)

miguelff:
@karmi, I understand your concern, and I agree we should make documentation clearer, and also agree about the importance of reading documentation.

[...]

PR #492, opened 2015-11-11
Delay pagination inclusion to first instantiation of Response

miguelff:
This change aims at serving as a first step for discussing alternative implementations to avoid https://github.com/elastic/elasticsearch-rails/issues/489

The approach is to delay the inclusion of the proper pagination module until the very first time a response object is created.

Before, inclusion of pagination modules was done from the outside of the `Response` class, accessing to the private `include` method in the class. Now, the inclusion is done from within the `Response` and memoized in the current thread.


miguelff:
Closing this as stale

PR #493, opened 2015-11-12
Uncommented gem 'sass'

sk187:
LoadError (cannot load such file -- sass) was due to commenting out sass gem.

### DIFF
--- a/elasticsearch-rails/lib/rails/templates/01-basic.rb
+++ b/elasticsearch-rails/lib/rails/templates/01-basic.rb
@@ -120,7 +120,7 @@ gem 'mocha', group: 'test', require: 'mocha/api'
 # ----- Remove CoffeeScript, Sass and "all that jazz" ---------------------------------------------
 
 comment_lines 'Gemfile', /gem 'coffee/
-comment_lines 'Gemfile', /gem 'sass/
+uncomment_lines 'Gemfile', /gem 'sass/
 comment_lines 'Gemfile', /gem 'uglifier/
 uncomment_lines 'Gemfile', /gem 'therubyracer/
 


sk187:
I signed the CLA twice :D? Does it take a while?

karmi:
@sk187 It's all right, the CLA has been signed. I'm not sure about "sass", would be cool to be able to run the app without it and all the asset compiling asociated with it.

sk187:
I just got a new computer and didn't have sass on my machine. When I tried to run the first example, it kept giving me a sass file error. 

Its because that sass gem is commented out. When I undid it and ran bundle install, everything worked out :D

karmi:
`sass` and `coffee` shouldn't be needed, that should be the point of commenting it out...

sk187:
Yea I don't think you need coffee but you do need sass. Otherwise you get a LoadError (cannot load such file -- sass).

karmi:
This is really weird, I've unistalled all the `sass*` gems on my machine, and run the template, and the application successfully starts and serves the pages:

```
$ gem list | grep sass

$ rails new searchapp --skip --skip-bundle --template ../lib/rails/templates/01-basic.rb 

...
        DONE    Starting the application.
...
Listening on localhost:3000, CTRL+C to stop


Started GET "/" for 127.0.0.1 at 2016-01-22 19:49:45 +0100
  ActiveRecord::SchemaMigration Load (0.2ms)  SELECT "schema_migrations".* FROM "schema_migrations"
Processing by ArticlesController#index as HTML
  Article Load (0.5ms)  SELECT "articles".* FROM "articles"
  Rendered articles/index.html.erb within layouts/application (11.4ms)
Completed 200 OK in 787ms (Views: 771.9ms | ActiveRecord: 0.8ms)
```
Can you please try again? I'm almost sure neither `sass` nor `coffee` should be needed, and I remember there were number of issues with these on Windows systems...

PR #500, opened 2015-11-26
Instance method `exists?`

Krule:
A convenience method to make it easier to verify whether index contains an object corresponding to instance id.


Krule:
Unnecessary as it can be done with `include Elasticsearch::Persistence::Repository::Find`

PR #502, opened 2015-12-03
Add the option to import data concurrently (using the rake task)

bernabas:
Added a `PARTITION` option to the import rake task

```
rake environment elasticsearch:import:model CLASS='Article' PARTITION=2 BY=id
```

this would create 2 processes and split the data set into 2 and run both processes in parallel ( i.e cutting import time by not letting active record be a bottleneck)

You would have to find the right partition number based on elasticsearch network throughput and memory settings) and/or adjust the settings per need.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,5 @@ _yardoc/
 coverage/
 rdoc/
 doc/
+.idea/
 Gemfile.lock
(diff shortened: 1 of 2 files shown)


bernabas:
@karmi can you review? :smile:

PR #504, opened 2015-12-04
Look for `from` and `size` in search body as well

Krule:
[Documentation says](https://www.elastic.co/guide/en/elasticsearch/reference/2.x/search-request-from-size.html):

> Though from and size can be set as request parameters, they can also be set within the search body.

This, maybe a bit to cautiously, _tries_ to implement it.

### DIFF
--- a/elasticsearch-model/lib/elasticsearch/model/response/pagination.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/response/pagination.rb
@@ -47,6 +47,8 @@ module Elasticsearch
             case
               when search.definition[:size]
                 search.definition[:size]
+              when search.definition[:body].try(:[], :size)
+                search.definition[:body][:size]
               else
                 __default_per_page
             end
@@ -58,6 +60,8 @@ module Elasticsearch
             case
               when search.definition[:from]
                 search.definition[:from]
+              when search.definition[:body].try(:[], :from)
+                search.definition[:body][:from]
               else
                 0
             end


Krule:
Closing for now until I have time to test it properly.

karmi:
@Krule Wondering what use case are you trying to support with the patch? I would need to dig deeper into the code, but I think that the current implementation _does_ in fact look into the body?

Krule:
@karmi If it does, that would be awesome as ATM I am using my own fork to make it work as I want it.

However, after I have closed this PR I took a look at the tests, I have found a few that appear to be ensuring the opposite:

https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-model/test/unit/response_pagination_kaminari_test.rb#L301-L308

https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-model/test/unit/response_pagination_kaminari_test.rb#L80-L87

As there are test ensuring this, I can only assume that it's intended behavior.

Krule:
@karmi Use case is simple as:

    POST _search
    {  "from": n,
       "size": m,
       ...
    }

encoded to JSON from hash:

    class ElasticsearchQuery
      ...
      def to_hash
        {
          from: from,
          size: size,
          query: { ... },
          ...
        }
      end
      ...
    end

Unfortunately, current implementation, as far as I could determine, did not look into body for `from` and `size` and I ended with default values. I will also look bit deeper as it would been strange if it's not supported, but I, so far, I have not been able to find the other reason.

karmi:
@Krule And how do you define the search in the `POST _search`? Normally, in a Rails context, you do something like this: https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-rails/lib/rails/templates/03-expert.rb#L213

Krule:
@karmi In my case was equivalent of:

    Article.search(ElasticsearchQuery.new(params).to_hash).results

with `from` and `size` as part of the hash. This did not work without the patch. However, once i changed it to equivalent of: 

    Article.search(ElasticsearchQuery.new(params).to_hash).page(params[:page]).results

everything worked as intended. 

This, for my current implementation, is sufficient.

However, elasticsearch does allows these parameters as part of the search body and I think pagination should also attempt to use them, provided they are not set as request params and not already implemented as part of pagination in this gem.

PR #517, opened 2016-01-05
Fix #dup for Elasticsearch::Model

jferris:
Previously, calling `#dup` for an `Elasticsearch::Model` instance would retain the original `__elasticsearch__` reference. Given the following example:

``` ruby
user = User.create!(name: "Will")
other = user.dup
other.update!(name: "Bill")
```

You'd end up with two references to "Will" in Elasticsearch, and none for "Bill," because the duplicate instance proxied to the original instance's attributes.

With this fix, each duplicate gets its own proxy, so attributes are saved correctly.

### DIFF
--- a/elasticsearch-model/lib/elasticsearch/model/proxy.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/proxy.rb
@@ -66,6 +66,15 @@ module Elasticsearch
         end
       end
 
+      # @overload dup
+      #
+      # Returns a copy of this object. Resets the __elasticsearch__ proxy so
+      # the duplicate will build its own proxy.
+      def initialize_dup(_)
+        @__elasticsearch__ = nil
+        super
+      end
+
       # Common module for the proxy classes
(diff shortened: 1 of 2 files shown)


jferris:
I signed the CLA, but I'm not sure how to get that check to re-run.

The build is failing during `bundle install` with `NoMethodError: undefined method `spec' for nil:NilClass`. It looks like some other pull requests are failing with that error, too, so I don't think I introduced that issue.

karmi:
@jferris I've re-run the CLA check manually, it's green, thanks! I'll run the test suite locally anyways when I get to it, so no worries!

karmi:
Hi Joe, many thanks for the patch, it makes perfect sense, hope it wasn't too much hassle to find... Apologies for the delay!, I've been too busy with other things.

The CI failures have been fixed in 55d765c, it was really tricky to figure out...

jferris:
@karmi excellent, thank you!

Issue #476, opened 2015-10-02
Kaminari Pagination with MultiModel Search


rkrdo:
This is related to #382 
Using the code from master, I'm trying to make a MultiModel Search:
(`Elasticsearch::Model.search('search term', [User, Admin]).page(params[:page]).records`)

And that works fine, however, when trying to use Kaminari's `paginate` helper with the **records** collection, I'm getting the following error:

```
NameError - undefined local variable or method `max_pages' for #<Elasticsearch::Model::Response::Records:0x007f99cd59b258>:
   () .../bundler/gems/elasticsearch-rails-ded203569208/elasticsearch-model/lib/elasticsearch/model/response/records.rb:58:in `method_missing'
  kaminari (0.16.3) lib/kaminari/models/page_scope_methods.rb:27:in `total_pages'
  kaminari (0.16.3) lib/kaminari/helpers/action_view_extension.rb:18:in `paginate'
```

I noticed that the Kaminari configuration methods are getting included in `Elasticsearch::Model::Response::Results` but not on `Elasticsearch::Model::Response::Records` (see [model/response/pagination.rb](https://github.com/miguelff/elasticsearch-rails/blob/v0.1.6/elasticsearch-model/lib/elasticsearch/model/response/pagination.rb#L19-L21)), so including the Kaminari methods in `Elasticsearch::Model::Response::Records` fixes my issue.

Is there a reason why those methods don't get included in  `Elasticsearch::Model::Records` ?

jonfer:
I have same problem

miguelff:
@rkrdo , @jonfer, is `kaminari` included in the Gemfile, before `elasticsearch-*`? (some context: https://github.com/elastic/elasticsearch-rails/issues/489, https://github.com/elastic/elasticsearch-rails/issues/487#issue-115380728)

jonfer:
@miguelff Yes

jonfer:
Ok, this commit in Master fixes the problem: https://github.com/elastic/elasticsearch-rails/commit/5f32e484a6d0458f26dc9acfa9dd5fed4e5d6453

rkrdo:
@miguelff yeah, `'kaminari` is included before `elasticsearch-model` and `elasticsearch-rails` in the Gemfile.

I'm using version `0.1.7` BTW, I noticed that `0.1.8` hasn't included the fix for this issue.
[...]

PR #452, opened 2015-08-10
Fix `next_page` method on Kaminari integration

ggcampinho:
Delegate `max_pages` to `Elasticsearch::Model::Response::Results` and `Elasticsearch::Model::Response::Response`, in order to  Kaminari method `next_page` work

The error:
```
> @records.next_page
NameError: undefined local variable or method `max_pages' for #<Elasticsearch::Model::Response::Records:0x007fb218974970>
from /Users/ggcampinho/.rvm/gems/ruby-2.2.2/gems/elasticsearch-model-0.1.7/lib/elasticsearch/model/response/records.rb:58:in `method_missing'
```

PR #524, opened 2016-01-22
[MODEL] Remove unnecessary code related to kaminari pagination

miguelff:
This PR removes unnecessary code in the pagination module, and ensures that the tests added as part of #452 actually check regressively the need for adding `max_pages` method to `Response::Results` and `Response::Records`.
### Context

In #452, the use of Kaminari's `paginate` helper was fixed by implementing the missing method in both the response results and the response records. 

However, the tests added as part of that PR were not actually checking any wrong behaviour, as proven by https://github.com/miguelff/elasticsearch-rails/commit/86b3590a1650a8a94558f6adcf5541022787534e, which removes the method declaration, while the test suite still  passes.

For those tests to proof a missing `max_pages` method was needed, an explicit inclusion of the `Kaminari::ConfigurationMethods` [needed to be removed](https://github.com/elastic/elasticsearch-rails/commit/9261bbddc5320163cb58b0ea31f946f64240f935#diff-ade00d41c79cf78b1678e1ea50010b7cL5).

Also, while investigating #476 I realised there's no longer needed to include `::Kaminari::Configuration::ClassMethods` in `Response::Results`, so [I removed it from the module](https://github.com/miguelff/elasticsearch-rails/commit/1f738cf3dfbd1c435c2a5257a9b871c8d84e9b75)

### DIFF
--- a/elasticsearch-model/lib/elasticsearch/model/response/pagination.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/response/pagination.rb
@@ -19,8 +19,8 @@ module Elasticsearch
             Elasticsearch::Model::Response::Results.__send__ :include, ::Kaminari::PageScopeMethods
             Elasticsearch::Model::Response::Records.__send__ :include, ::Kaminari::PageScopeMethods
 
-            Elasticsearch::Model::Response::Results.__send__ :delegate, :limit_value, :offset_value, :total_count, :max_pages, to: :response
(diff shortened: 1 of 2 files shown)


miguelff:
The build fails when running the PR, however https://travis-ci.org/miguelff/elasticsearch-rails is green, and the branch is a FF of master.

karmi:
Hi Miguel, thanks for the patch, from what I saw it makes sense, will run the tests locally and have a look at it in the next couple of days...

miguelff:
Closing this as stale

Issue #199, opened 2014-08-12
Add support for saving multiple items in bulk from Repository#save


karmi:
Currently, the repository can save objects individually, with no convenient interface for bulk processing.

A polymorphic interface should be supported, where the `Repository#save` method would take one or more items, and save the objects either individually or in bulk:

``` ruby
note1 = Note.new id: 1, text: 'N1'
note2 = Note.new id: 2, text: 'N2'

repository.save(note1)
# -> Individual save

repository.save([note1, note2])
repository.save(note1, note2)
# -> Bulk
```

PR #527, opened 2016-02-04
Bulk save support for repository

vhyza:
Hi @karmi,

I tried to implement bulk save support for repository pattern (as mentioned in issue https://github.com/elastic/elasticsearch-rails/issues/199).

I'm afraid we can't provide `repository.save(note1, note2)` signature as mentioned in issue, because of optional `options` hash. AFAIK we can pass `Hash` to the `repository#save` method, so I think we can't distinguish:

``` ruby
note1 = { id: 123, content: '' }
note2 = { id: 123, content: '' }
options = {}

repository.save(note1, note2, options)
```

and

``` ruby
repository.save(note1, options)
```

If we are sure, that document is not Hash, we can use something like `def save(*docs, **options)` but as I wrote before. I'm not sure if we can do something like that if document can be `Hash`.

So for now possible use is:

``` ruby
note1 = { id: 123, content: '' }
note2 = { id: 123, content: '' }

repository.save(note1)
# -> individual save

repository.save([note1, note2])
# -> bulk save
```

My second concern is, that `repository#save` now returns two format of responses based on client method call (`index` vs `bulk`). What do you think?

### DIFF
--- a/elasticsearch-persistence/lib/elasticsearch/persistence/repository/naming.rb
+++ b/elasticsearch-persistence/lib/elasticsearch/persistence/repository/naming.rb
@@ -108,6 +108,23 @@ module Elasticsearch
         def __extract_id_from_document(document)
           document.delete(:id) || document.delete('id') || document.delete(:_id) || document.delete('_id')
         end
+
+        # Extract an attribute from the document (assuming Hash or Hash-like object)
+        #
+        # @note Calling this method will *remove* the `attribute` key from the passed object.
+        #
+        # @example
+        #     options = { title: 'Test', id: 'abc123', _parent: 'parent' }
+        #     repository.__extract_attribute_from_document(options, :_parent)
+        #     # => "parent"
+        #     options
+        #     # => { title: 'Test', id: 'abc123' }
+        #
+        # @api private
+        #
+        def __extract_attribute_from_document(document, attribute)
+          document.delete(attribute.to_sym) || document.delete(attribute.to_s)
+        end
       end
 
     end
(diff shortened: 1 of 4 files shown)


triptec:
@karmi this would be really nice to get bulk for the repo. Would this be merged if we can fix the tests?

triptec:
I only find the faling build that's 8 months old, you have a link? The build that i started failed on Gem::InstallError: mime-types-data requires Ruby version >= 2.0. (https://travis-ci.org/elastic/elasticsearch-rails/jobs/141333264#L781)

vhyza:
Here is build 4 hours old: https://travis-ci.org/elastic/elasticsearch-rails/builds/141288956

triptec:
Ah, okay, how do we fix this then?

vhyza:
It is quite a while I pushed this PR so I need to setup the environment and try tests locally...

triptec:
Any news on this?

triptec:
@karmi this is something that would be extremely useful, it it possible you could take a look and see what might be at fault?

vhyza:
@triptec sorry, I'm mostly offline these days, but I think, you can add `save`, `__save_one`, `__save_many` located in https://github.com/vhyza/elasticsearch-rails/blob/2c7b7b38416a03ecc103c7979ac1fda0ea6f667b/elasticsearch-persistence/lib/elasticsearch/persistence/repository/store.rb directly into your repository implementation; it'll be probably quicker based on the response times in this issue :)

MarcLapierre:
Is this feature still actively being worked on?

PR #530, opened 2016-02-17
Set the index name as block for Elasticsearch::Persistence::Repository

otzy007:
Set the index name as a block similar with Elasticsearch::Model https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-model/lib/elasticsearch/model/naming.rb#L21

This way the index name can be re-evaluated at every call.


estolfo:
Closing as this is no longer relevant.

PR #531, opened 2016-02-18
Indexing settings: don't modify input hash

urbanautomaton:
Hi,

At the moment, if a client passes a frozen hash to `Elasticsearch::Model::Indexing.settings`, a `RuntimeError` ensues, objecting that the hash has been modified. This is because the first call to .settings [initializes a `Settings` instance using the provided hash](https://github.com/Futurelearn/elasticsearch-rails/blob/489ab24998a87a0323c3085e29aad14367bfaf89/elasticsearch-model/lib/elasticsearch/model/indexing.rb#L191), then immediately [merges the same hash with itself](https://github.com/Futurelearn/elasticsearch-rails/blob/489ab24998a87a0323c3085e29aad14367bfaf89/elasticsearch-model/lib/elasticsearch/model/indexing.rb#L193).

This patch addresses the issue by initializing the Settings instance with a new, empty hash, and applying all updates to this.

As part of this PR I've made two further changes. When writing the test for my patch I noticed that there was some order-dependent behaviour in the `Indexing` tests, which meant that I would have needed to assert the presence of settings that weren't applied in my test. I've addressed this in https://github.com/elastic/elasticsearch-rails/commit/350b7eaef8268013f787c632495c2ab6dfa8e416 by using anonymous classes instead of the shared dummy classes the tests previously used.

I've also slightly refactored the `.settings` method, primarily for readability. I've kept this as a separate commit, though, so I'm happy to remove this change if you'd prefer.

Cheers,
Simon

---

Edit: I've signed the CLA, but I'm not sure what to do about the failing build, which appears unrelated to my changes. The tests pass locally, and I'm not able to reproduce the failure.

### DIFF
--- a/elasticsearch-model/lib/elasticsearch/model/indexing.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/indexing.rb
@@ -22,6 +22,10 @@ module Elasticsearch
           @settings = settings
         end
 
+        def update(additional_settings={})
+          @settings.update(additional_settings)
+        end
+
         def to_hash
           @settings
         end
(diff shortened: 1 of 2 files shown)


urbanautomaton:
Hi - just wanted to check in on this and see if there's anything I need to do that could help it be ready for merge.

My employer and I have both signed the CLA - I'm not sure how to get the automated check for that to pass, however.

karmi:
Hi @urbanautomaton, thanks for the patch! My apologies for the silence here. I agree it's a great change, and thank you for having the energy to fix the tests as well?

I can see you have signed the individual CLA in February already, our systems sometimes haven't received all the signatures, and it happened to you as well. I've added your signature manually to our systems and triggered the check, and all is green.

Please _do ping me_ next week if I fail to attend to your contribution again!

urbanautomaton:
I've lost track of why this patch was important to me in the first place and it no longer applies cleanly, so I'm going to close it. (This is not a complaint at all btw! 😄)

PR #549, opened 2016-03-23
Update README.md

sk187:
Tell users the significance of dynamic mapping


karmi:
Hi @sk187, I understand this might be really frustrating thing to "discover". On the other hand, I wouldn't like to overload the README with "notes" for this and that -- the "dynamic" nature of Elasticsearch is pretty typical of it, and the user should be aware of that... We can of course put such note into the RDoc for corresponding methods in the `elasticsearch-model` and `elasticsearch-api` gems.

karmi:
Closing due to silence, please reopen or ping me for further info!

Issue #554, opened 2016-04-08
Rake elasticsearch::import::model Progressbar and scopes


caomania:
When an scope is provided to the `elasticsearch::import::model` rake task. This scope is not applied to the `count` total used by `ANSI::Progressbar`. This causes the progress bar visualisation to be inaccurate.

From [elasticsearch-rails/elasticsearch-rails/lib/elasticsearch/rails/tasks/import.rb](https://github.com/elastic/elasticsearch-rails/blob/a714d5931182501531d2ce6b767bcb3c97ce56a7/elasticsearch-rails/lib/elasticsearch/rails/tasks/import.rb#L50-L51)

``` ruby
total  = klass.count rescue nil
pbar   = ANSI::Progressbar.new(klass.to_s, total) rescue nil
```

PR #555, opened 2016-04-08
Apply scope to total count query

caomania:
[#554]


caomania:
I don't see any test harness for `rake` tasks currently in place. For this reason I haven't included an explicit test for this change.

PR #556, opened 2016-04-11
Make find_in_batches work with rails 3

carlallen:
`extract!` returns a hash with nil for the values of the extracted keys in rails 3, this does bad things when you try to merge back in the values.


estolfo:
@carlallen Can you please add a test and provide an example of the problem this is solving? Thanks

estolfo:
Closing as there's no response.

PR #565, opened 2016-05-07
Fix Elasticsearch-model README typo under Importing the Data

sagarpatel8384:


karmi:
Not a native English speaker :), but I think it should be either `is importing your data` or `is to import your data`. I guess I went for the former to decrease the amount of `to` :)

sagarpatel8384:
No problem :) Both `is to import your data` and `is import your data` are grammatically correct.

PR #574, opened 2016-05-29
Update elasticsearch-model/README.md  [ci skip]

bogdanvlviv:
Added mention about elasticsearch-dsl in elasticsearch-model/README.md


karmi:
Thanks!, that's nice to thing to add to the model README -- I've reworded it a bit and added an example.

PR #582, opened 2016-06-27
add edge-ngram elastic search example

equivalent:
example of https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-edgengram-tokenizer.html 

this is supper helpful to understand not only edge-ngrams, but how to actually define own `analysis` with elasticsearch-rails

### DIFF
--- a/elasticsearch-model/examples/activerecord_mapping_edge_ngram.rb
+++ b/elasticsearch-model/examples/activerecord_mapping_edge_ngram.rb
@@ -1,0 +1,100 @@
+require 'ansi'
+require 'sqlite3'
+require 'active_record'
+require 'elasticsearch/model'
+
+ActiveRecord::Base.logger = ActiveSupport::Logger.new(STDOUT)
+ActiveRecord::Base.establish_connection( adapter: 'sqlite3', database: ":memory:" )
+
+ActiveRecord::Schema.define(version: 1) do
+  create_table :articles do |t|
+    t.string :title
+    t.date   :published_at
+    t.timestamps
+  end
+end
+
+class Article < ActiveRecord::Base
+  include Elasticsearch::Model
+  include Elasticsearch::Model::Callbacks
+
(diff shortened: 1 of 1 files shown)


equivalent:
@karmi I've tried to click several times on the ` An individual contributor` and Sign Agreement but it looks it's not doing anything

equivalent:
tests are failing due to `Make sure that `gem install mime-types-data -v '3.2016.0521'` succeeds before`  nothing to do with the code change,

equivalent:
@karmi any feedback on this PR ?

karmi:
Hi Tomas, apologies for this ridiculous delay; I haven't been able to find time to process patches and respond to questions in this repository for a long time.

I've re-worded the commit message a bit, and I've translated the analysis output you had in comments into code, using the "Indices Analyze" API, in hope that it will add some didactic value to people.

When I was considering the example just now, I was wondering if we should have exactly this as an example for "home-grown" autocomplete -- people should use the dedicated [_Completion Suggester_](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-suggesters-completion.html) for that if at all possible; there's a Ruby example in `activerecord_mapping_completion.rb`.

However, I think the example still has value, and it's actually good that it is feature-wise very similar to the "completion" example. I'm thinking about actually expanding it a little bit, maybe by adding a different ngram filter, which would allow "infix" (inside-the-words) searches.

equivalent:
Hi Karmi,  no worries. 

Yeah it's up to you really I've got this piece of code in my Gist so people can still find it if they need :) 
So feel free to change it / close it/ rewrite it. During the time I was writing this for my employer I struggled to find similar example so had to come up with own one. So thought it would be helpful to have it here + it it's a good way to get feedback by a Pro on this topic. 

čus

equivalent:
@karmi awesome !! :) thank you

PR #583, opened 2016-06-27
generate  Paperclip urls with data provided via Elasticsearch

equivalent:
instructions how  to render Paperclip urls without ActiveRecord calls

Based on this article: http://www.eq8.eu/blogs/26-rendering-paperclip-url-via-elasticsearch-without-making-calls-to-activerecord-db

### DIFF
--- a/elasticsearch-model/examples/activerecord_mapping_rendering_paperclip_attachment_urls.rb
+++ b/elasticsearch-model/examples/activerecord_mapping_rendering_paperclip_attachment_urls.rb
@@ -1,0 +1,158 @@
+require 'ansi'
+require 'sqlite3'
+require 'active_record'
+require 'elasticsearch/model'
+require 'paperclip'
+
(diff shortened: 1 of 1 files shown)


elasticsearch-model/examples/activerecord_mapping_rendering_paperclip_attachment_urls.rb
```diff
@@ -0,0 +1,158 @@
...
+  end
+end
+
+begin
+  class Image < ActiveRecord::Base
+    belongs_to :property, touch: true
+    has_attached_file :attachment, styles: { thumb: '100x100>', screen: '1024x1024' }
+  end
+rescue => e
+  puts e.inspect.ansi(:bold, :red)
```
equivalent: Paparclip is nasty, in some cases it may not be loaded properly and the `has_attached_file` will report `NoMethodError`

PR #585, opened 2016-06-30
Proxy object destroys inheritance feature

kayakyakr:
The tests worked great when the objects just had `Elasticsearch::Model::Naming` methods included, but that's not how Elasticsearch is used in real life. In real life, the index_name and document_name are accessed through a proxy, and the tree search was not considering that. This adds tests to confirm the infinite loop that this caused, and then fixes it by checking against the proxy object's target.

### DIFF
--- a/elasticsearch-model/lib/elasticsearch/model/naming.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/naming.rb
@@ -77,7 +77,7 @@ module Elasticsearch
 
             if Elasticsearch::Model.settings[:inheritance_enabled]
               self.ancestors.each do |klass|
-                next if klass == self
+                next if klass == self || self.respond_to?(:target) && klass == self.target
                 break if value = klass.respond_to?(prop) && klass.send(prop)
               end
             end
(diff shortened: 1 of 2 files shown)


kayakyakr:
Errors are unrelated to code changes.

tmandke:
Can this get merged soon. We are facing the same problem.

jsmestad:
This would be great to get merged!

karmi:
@jsmestad , @tmandke , @kayakyakr , sorry about the silence here... I still need to focus on other responsibilities currently, but we're scheduling some significant effort for June and onwards to re-focus on the `elasticsearch-rails` library again, and give it the much needed attention.

estolfo:
@kayakyakr Thanks very much for this pull request. Is there any chance you can create a test that uses the actual library code? I'm afraid that if anything changes in the code base, this test will erroneously continue to pass because of the use of the `::ProxiedModel`

kayakyakr:
I made some comments in the spec why that wasn't possible: testing against the model itself causes inheritance loops. I'm no longer working on a project that uses this, and so no longer have time set aside in my workday to work on finding a different workaround for this.

PR #586, opened 2016-06-30
generate  Paperclip urls with data provided via Elastic search

equivalent:
instructions how to render Paperclip urls without ActiveRecord calls

Based on this article: http://www.eq8.eu/blogs/26-rendering-paperclip-url-via-elasticsearch-without-making-calls-to-activerecord-db

...closed the [old PR](https://github.com/elastic/elasticsearch-rails/pull/583) and reopening it in order to retriger CLA github hook

### DIFF
--- a/elasticsearch-model/examples/activerecord_mapping_rendering_paperclip_attachment_urls.rb
+++ b/elasticsearch-model/examples/activerecord_mapping_rendering_paperclip_attachment_urls.rb
@@ -1,0 +1,158 @@
+require 'ansi'
+require 'sqlite3'
+require 'active_record'
+require 'elasticsearch/model'
+require 'paperclip'
+
+require "paperclip/railtie"
+Paperclip::Railtie.insert
+
+ActiveRecord::Base.logger = ActiveSupport::Logger.new(STDOUT)
+ActiveRecord::Base.establish_connection( adapter: 'sqlite3', database: ":memory:" )
+
+ActiveRecord::Schema.define(version: 1) do
+  create_table :properties do |t|
+    t.string :title
+    t.string :description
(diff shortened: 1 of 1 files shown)


equivalent:
@karmi any feedback on this PR ?

karmi:
Hi Tomas, apologies for the lack of feedback on this patch as well!

I've been thinking about it, and while I understand the general motivation of what you're trying to demonstrate, I'm worried that the focus of this example is not clear enough.

First, I'm worried that people see "paperclip" in the example name and get an idea that it will show how to handle attachments, only to be frustrated later, because it doesn't :)

Second, if all we're trying to demonstrate here is that it's frequently best to work with the data loaded from Elasticsearch, the `activerecord_associations.rb` should be the place to do that. I've expanded this example in the attached commit (a086ded). Can you look at it and ping me back if you think it's more or less what you had in mind? Maybe we can either further expand the "associaton" example, or work on the "paperclip" example to cover more features in a clear way.

equivalent:
hi @karmi ,

sorry I didn't get back sooner, 

basically my aim  is to demonstrate how to  render the paperclip (or Carrierwave doesn't really matter) image url without touching relational database. 
My only motivation for this PR was that I wrote that blog post where I demonstrate how to do this and I told myself: I may as well just create PR here :) 

back to your points:

1 yes, fully agree. It's  hard to explain that this is just for "view" only  in one file. Therefore for sake not to cause confusion I'm ok to close this PR

2 pretty much yes, I understand you are trying to explain general principle so that people understand how to approach this type of data  where as my PR was just Copy - Paste - Edit solution for the lazy ones :) 

so I'm ok to close this PR

Issue #526, opened 2016-02-03
Elasticsearch::Persistence::Model::before_update: access to the values to be updated


jogaco:
In order to update a dependent pre-calculated field, I would need to access the new values in the before_update callback.

Related to #514

Issue #514, opened 2015-12-28
Elasticsearch::Persistence::Model::before_update: attributes assigned within are not persisted


jogaco:
Example:

```
class Artist
  include Elasticsearch::Persistence::Model
...
  attribute :profile
  before_update :non_empty_profile

  def non_empty_profile
    if self.profile.blank?
      self.profile = 'Profile'
    end
  end
end
```

Even though the code is run for instance in `@artist.update(artist_params, refresh: true)`
the assigned value to profile is not persisted

PR #599, opened 2016-07-21
[PERSISTENCE] Fix for #514, #526: Allow before_update callback to access new values…

jogaco:
… of attributes and alter them


jogaco:
CLA just confirmed, after the PR

PR #602, opened 2016-07-22
add Rails Active Job async example

equivalent:


equivalent:
CI is failing due to unrelated reason

```
An error occurred while installing activesupport (5.0.0), and Bundler cannot
continue.
Make sure that `gem install activesupport -v '5.0.0'` succeeds before bundling.
```

Issue #603, opened 2016-07-22
Elasticsearch::Persistence ActiveRecord pattern: after_find callback not called


jogaco:
This callback is good to customize behaviour after deserialization. Trouble is it is not called.

karmi:
Hi @jogaco, sorry for the long delay on this one... I'll need some time to wrap my head around it. I'll ping you with questions when I have them.

jogaco:
As reported in the PR, this is actually not a feature, but a non-implemented callback
https://github.com/elastic/elasticsearch-rails/blob/7aa03e87b7fee71007ff7aa1a6fb452de298587e/elasticsearch-persistence/lib/elasticsearch/persistence/model.rb#L38

jogaco:
Any progress with this @karmi please? This is a show stopper.

For elasticsearch-persistence Repository Pattern, there is customizable deserialization:

https://github.com/elastic/elasticsearch-rails/blame/109c155df6c8b8989aa935bc1fe163a1452a2b4e/elasticsearch-persistence/README.md#L144-L148

For elasticsearch-persistence ActiveRecord Pattern, there is a promise for it in the defined Model callbacks. Only trouble is *it is not implemented*. My PR is just one line of code that **fulfils the promised declared callback**.

https://github.com/jogaco/elasticsearch-rails/blob/bce287ed35afb855e93eeebc3c5a7c363523c164/elasticsearch-persistence/lib/elasticsearch/persistence/model.rb#L119-L121

If there is need for additional modifications in code, or anything I can do to facilitate the PR mersing, please let me know and I'll do my best.

postmodem:
Having this fix merged would be very beneficial to me as well. Any word on merging?

estolfo:
Closing as the ActiveRecord persistence pattern has been deprecated.

PR #604, opened 2016-07-22
[PERSISTENCE] Fix for #603 after_find callback

jogaco:
Fix for #603 after find callback


ryansch:
This appears to include the changes from #599 as well.

jogaco:
Closing and new pull-request with only relevant commit. Thanks!

Issue #610, opened 2016-08-03
[PERSISTENCE] make change of updated_at optional


jogaco:
This could be passed in options:

`item.update({title: 'title'}, {touch: false})`

Most useful for massive updates to fix field values.

PR #611, opened 2016-08-03
[PERSISTENCE] Fix for #610: make change of update at optional

jogaco:


ryansch:
This appears to also include the changes from #604.

jogaco:
New PR #653 with just the changes for #610

PR #616, opened 2016-08-14
Use correct method in repository count example.

robwilliams:
The Repository 'count' example was using 'search' as the method name
instead of 'count'.


estolfo:
Updated in this commit: https://github.com/elastic/elasticsearch-rails/commit/b3237ee04225c3c6f60f8767396364fd9f24240d

Issue #608, opened 2016-08-01
ActiveRecord collection not (obviously) sorted by score


alexceder:
All this came about when updating from **Rails 4.2** to **Rails 5**.

I have hade some trouble with Elasticsearch not returning ActiveRecord models in order of score.

Reviewing the response, I can see that they are **correctly returned in order by score** so that:

```
response.map(&:id)
# => ["13", "12"]
```

However this is **not what I want**:

```
response.records.map(&:id)
# => [12, 13]
```

And this is where you might thing.. well you need to call `to_a` on the records. And yes, that works.

```
response.records.to_a.map(&:id)
# => [13, 12]
```

Just as the README states, it runs the corresponding SQL with only a `WHILE` statement. This confused me, where do you sort!?

https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-model/README.md#search-results-as-database-records

The sorting is done when explicitly calling `to_a`. This is of course fine, but this isn't the big thing that confused me or where my errors appeared.

I first used this solution, but I thought it was rather ugly. This way I get the records sorted properly from the database, on primary key and not undefined behaviour.

```
response.records.order("FIELD (`table_name`.`primary_key`, 13, 12)").to_a.map(&:id)
# => [13, 12]
```

But scrap that, I don't want to write custom SQL for which should happen automagically.

So why am I writing this loooong wall of text.. Well, it seems like Rails has changed how it uses implicit convertions.. or something.. so.. This is probably where something happens (or more likely doesn't happen):

```
# Wrong ordering
render json: records
```

Because if I explicitly convert to array, all is well.

```
# Correct ordering
render json: records.to_a
```

This might be more of a **gotcha**, but it was a pain for me, and might be for others.

Is there a better way of doing this? Some other intended use? Is this dumb? Could the "default" [order](https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-model/lib/elasticsearch/model/adapters/active_record.rb#L50-L67) be done like I did it? Ideas? Dumb?

johanlunds:
Possibly related:

* https://github.com/elastic/elasticsearch-rails/issues/258
* https://github.com/elastic/elasticsearch-rails/issues/594
* https://github.com/elastic/elasticsearch-rails/issues/546
* https://github.com/elastic/elasticsearch-rails/issues/169
* https://github.com/elastic/elasticsearch-rails/issues/568

dimroc:
[...]

PR #618, opened 2016-08-17
[Model] Fix ActiveRecord sorted collection

inkstak:
This is a fix for #608.
It also introduces tests against ActiveRecord 5.0.

The fix about sorted records is cover by a test, but there are still 10 errors on other subjects when running tests with AR 5.
I'didn't run unit tests as they are already failing.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -35,6 +35,8 @@ namespace :bundle do
     sh "bundle install --gemfile #{__current__.join('elasticsearch-model/gemfiles')}/3.0.gemfile"
     puts '-'*80
     sh "bundle install --gemfile #{__current__.join('elasticsearch-model/gemfiles')}/4.0.gemfile"
+    puts '-'*80
+    sh "bundle install --gemfile #{__current__.join('elasticsearch-model/gemfiles')}/5.0.gemfile"
   end
 
   desc "Remove Gemfile.lock in all subprojects"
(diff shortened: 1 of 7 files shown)


nickrivadeneira:
@inkstak You'll need to sign the contributor agreement: https://www.elastic.co/contributor-agreement

Would love for this to be merged in.

sfcgeorge:
Confirm this fix works for me, without it result order is all over the place. Had to monkeypatch it in as Rails 5 has been out long enough.

connorshea:
@nickrivadeneira any updates on this?

nickrivadeneira:
@connorshea I'm not a elasticsearch member unfortunately, I was just making sure that instak signed the agreement so this could be merged in when the elasticsearch team looks at it. This repo seems to be dead though :( The person we need is @karmi

inkstak:
I confirm I signed the agreement. The github check is not up to date.

karmi:
@inkstak, you were correct, the check has indeed been out of sync -- it's corrected now.

@nickrivadeneira, it's not dead, I just have a hard time keeping up with all the gems across the two repos. I'll pull the patch and verify locally and merge.

nickrivadeneira:
@karmi Makes sense. Thanks for looking into this one

karmi:
Hi, I've merged the patch into `master`, many thanks for that, and apologies again for the delay. I'm not sure when I'll get to a release, but if you want a patch release with the change, I can do it anytime.

(So, one reason why processing PRs is not a simple "pull, run tests, rewrite commit message, push" is that I needed fix [all this](https://github.com/elastic/elasticsearch-rails/compare/94d0b28...master) while doing the pull, so the tests are green and we can add more patches on top of the branch without fear...)

inkstak:
Thanks 👍

PR #621, opened 2016-08-25
[RAILS] Fixed 03-expert.rb set tracer only in dev env

timoschilling:
The variable `tracer` is only defined in Rails development env, so we only can set them in that env.


estolfo:
thanks @timoschilling !

PR #628, opened 2016-09-28
[MODEL] Scope #records to current collection scope

nickrivadeneira:
When calling `#search` on a scoped collection, `#records` would ignore
that scope and query the full model collection. This change maintains
the scope `#search` is called on when calling `#records`. It does not
filter the elasticsearch query, so `#results` is unaffected.


nickrivadeneira:
Closing since this hasn't gotten much attention and I'm not sure if it's an issue w/ the latest version or not.

PR #642, opened 2016-11-17
Fix error in Async Indexing example

adarsh:
Reason for Change
=================
* The"Asynchronous Callbacks" section of the `README.md` contains an example Sidekiq job class.
* I kept getting a "undefined method 'as_indexed_json'", even though I had mixed in `Elasticsearch::Model` which was just puzzling.
* This example class contains an error:

```ruby
Client.index blah: 'blah', body: record.as_indexed_json
```

should be

```ruby
Client.index blah: 'blah', body: record.__elasticsearch__.as_indexed_json
```

Changes
=======
* Fix the error in the `README.md` file.

### DIFF
--- a/elasticsearch-model/README.md
+++ b/elasticsearch-model/README.md
@@ -518,7 +518,7 @@ class Indexer
     case operation.to_s
       when /index/
         record = Article.find(record_id)
-        Client.index  index: 'articles', type: 'article', id: record.id, body: record.as_indexed_json
+        Client.index  index: 'articles', type: 'article', id: record.id, body: record.__elasticsearch__.as_indexed_json
       when /delete/
         Client.delete index: 'articles', type: 'article', id: record_id
       else raise ArgumentError, "Unknown operation '#{operation}'"


ryansch:
This would prevent the caller from overriding `#as_indexed_json` on the model.  We use this pattern in every one of our models that include `Elasticsearch::Model`.

adarsh:
Perhaps I'm being unclear, my apologies.

I'm saying "mixing in `Elasticsearch::Model` does not add in the `.as_indexed_json` method to the class, as this example is implying. You need to invoke the client via `__elasticsearch__` for the method to be available."

Here's an example:

```ruby
irb(main):006:0> class Example; include Elasticsearch::Model; end
Example < Object
irb(main):007:0> Example.new.respond_to?(:as_indexed_json)
false
irb(main):008:0> Example.new.__elasticsearch__.respond_to?(:as_indexed_json)
true
```

> This would prevent the caller from overriding #as_indexed_json on the model. We use this pattern in every one of our models that include Elasticsearch::Model.

I am probably missing something, but I'm not clear on how you could override a method which does not exist?

ryansch:
You're absolutely right.  es-model is actually proxying from the _proxy_ to the user defined class (which is terribly confusing).

It appears that the right way to call `as_indexed_json` is `__elasticsearch__.as_indexed_json`.  That call signature will still respect an overridden definition.

```ruby
irb(main):001:0>  class Example; include Elasticsearch::Model; def as_indexed_json(options={}); 'foobar'; end; end
=> :as_indexed_json
irb(main):002:0> ex = Example.new
=> #<Example:0x0000000de4a7e8>
irb(main):003:0> ex.__elasticsearch__.as_indexed_json
=> "foobar"
```

adarsh:
@ryansch makes sense. Any edits before this can be merged?

adarsh:
This is starting to get stale - any blockers before this can be merged? (cc @ryansch).

ryansch:
@adarsh You'll have to ping someone with commit access.  I can't merge your PR.

adarsh:
Thanks @ryansch!

This correction to the `README.md` is ready to merge! Any blockers I can resolve before this can be merged?

(Paging @karmi or @tylerjl)

karmi:
Hi, yes, sorry for the ridiculous delay... I always wanted to check the situation before blindly merging the README change, because the original code — `record.as_indexed_json` — _should_ work...

ryansch:
@karmi See the first three comments on this PR

estolfo:
I believe this is already in the code base as of this commit https://github.com/elastic/elasticsearch-rails/commit/d5247aa1c98b24b2b5aaedb0f16cb0e71a0e30a3#diff-6b0d0617cb76ffd2503dd4708277bceb

PR #649, opened 2016-12-19
Move data type from string to text to be compatibel with ES 5.1

jalberto:


karmi:
Hi @jalberto, thanks for the patch! I'm afraid the codebase, and especially the _examples_ will need much more updates to be compatible with 5.x, and I have to set up some branches and versioning scheme for that. Would you be up for going over more places in the codebase? I've run the integration tests for `elasticsearch-model` against 5.x now, and there's just errors related to `multi_field`, but I think there'll be more.

jalberto:
@karmi I started to do so but I found several issues relates to mappings (the text/keyword thing & the properties keyword is not supported) I am afraid there will be more changes as you point out.

I am afraid this will require a deeper change so this not happens again in next versions (probably 1st step will be to remove virtus and add a DSL for ActiveRecord pattern). 
Thinking in all this changes I wonder if is not better just to deprecate ActiveRecord pattern in favour of repository pattern (as it allows more flexibility for changes)

I find the AR pattern confusing & error prune as ES cannot be used as an regular AR DB (thinking in PG, mongo, etc) as you cannot have migrations, some changes requires reindexing, etc

Another option would e to implement the persistence layer using dry-rb, it's flexible enough to use for both, the repository adn AR patterns.

karmi:
So, I've spent couple of days last week fixing the [Rails application templates](https://github.com/elastic/elasticsearch-rails/tree/master/elasticsearch-rails/lib/rails/templates) for the ActiveRecord integration, and while it's time-consuming to verify and double-check everything, the changes are [relatively straightforward](https://github.com/elastic/elasticsearch-rails/commit/5c692b74f8cd5d1fb1bbca4670d726bbf850ef4b).

I guess it will be a bit harder for the `elasticsearch-persistence` [examples](https://github.com/elastic/elasticsearch-rails/tree/master/elasticsearch-persistence/examples), but mostly similar.

This change, from `string` to `text` is pretty straightforward, and I'll have to do a similar change in the `indexes` method.

> Thinking in all this changes I wonder if is not better just to deprecate ActiveRecord pattern in favour of repository pattern (as it allows more flexibility for changes).

It's a very interesting idea. In my experience, the _ActiveRecord_ pattern _is_ indeed confusing and error-prone, not so much because of the Elasticsearch layer, but because of subtle problems and insufficient implementations in the `elasticsearch-persistence` gem. When I consider the issues and patches for this project, most are connected to the ActiveRecord pattern in one way or another, eating resources which could be spent on other parts of the library. And I definitely agree that the _Repository_ pattern is much more easy to develop in the library, and arguably easier to use in application code. I'll think about it.

jalberto:
Thansk for your work @karmi 

I already moved my projects to repository pattern, it required a small refactor, but was quite stright forward.
I suggest to promote the repository pattern as the recommended one, the lexx people using the AR pattern the better :)

> is indeed confusing and error-prone, not so much because of the Elasticsearch layer

Is not because the layer, but in how different ES. Providing an AR pattern that lacks some of the behavious AR users expect is very error prone.

Issue #450, opened 2015-08-03
`update` does not use setters before persisting to ES


natematykiewicz:
A `create` will coerce values using Virtus because `initialize` will coerce, and `save` will use `self.attributes`.

`update` does not read from `self.attributes`, but instead just sets the values in ES, and then updates the object's `self.attributes` after the save successfully occurs. This means that any custom coercion done with Virtus doesn't take effect when using `update`.

(reference: https://github.com/elastic/elasticsearch-rails/blob/v0.1.7/elasticsearch-persistence/lib/elasticsearch/persistence/model/store.rb#L115)

natematykiewicz:
This problem is a lot worse than I originally realized because it doesn't take ANY setters into account. It just blindly sets the values in ES to what the user passed in, regardless of whether there's an attribute defined for it or not.

This is a real problem, especially when using has_secure_password.
Calling `member.update(:password => 'foo')` will store, the unencrypted password field and value in ES, when it should be storing password_digest.

The update method should only allow you to change defined attributes, and be smart enough to realize that after updating attributes, the password_digest attribute is now dirty.

knshiro:
+1
Due to the same thing it does not check any validation either.

k-rudy:
+1 guys, the issue seems to be there for a long time already and is pretty important, could you please update us if anyone is working on that and what's the ETA for the fix.

estolfo:
Please see my comment [here](https://github.com/elastic/elasticsearch-rails/pull/658#issuecomment-419400259). We have deprecated and removed the `ActiveRecord` pattern in the elasticsearch-persistence gem.

PR #658, opened 2017-01-18
Validate updates

dgmora:
As pointed out in https://github.com/elastic/elasticsearch-rails/issues/450 and https://github.com/elastic/elasticsearch-rails/pull/289, the validation doesn't work when you make an update that makes a field invalid.

My knowledge in elasticsearch is very limited, so I am not sure if updating in [line 115](https://github.com/elastic/elasticsearch-rails/compare/master...dgmora:validate-updates?expand=1#diff-afd98d0d6bf4059de9d5b4ef2a0996a4R115) works well with the attributes being passed directly in [line 126](https://github.com/elastic/elasticsearch-rails/compare/master...dgmora:validate-updates?expand=1#diff-afd98d0d6bf4059de9d5b4ef2a0996a4R126). You **might** want to pass `self.attributes` to update instead.

### DIFF
--- a/elasticsearch-persistence/lib/elasticsearch/persistence/model/store.rb
+++ b/elasticsearch-persistence/lib/elasticsearch/persistence/model/store.rb
@@ -112,6 +112,7 @@ module Elasticsearch
           # @return [Hash] The Elasticsearch response as a Hash
           #
           def update(attributes={}, options={})
+            attributes = self.attributes.merge(attributes)
             unless options.delete(:validate) == false
               return false unless valid?
             end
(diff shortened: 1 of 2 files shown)


hfenton:
Any priority on getting this to pass CI and merged? IMO, this is a huge flaw in the `update_attributes` when using the `validate:` option.

elasticsearch-persistence/lib/elasticsearch/persistence/model/store.rb
```diff
@@ -112,6 +112,7 @@ def destroy(options={})
           # @return [Hash] The Elasticsearch response as a Hash
           #
           def update(attributes={}, options={})
+            attributes = self.attributes.merge(attributes)
```
hfenton: This should be `self.attributes = self.attributes.merge(attributes)`

estolfo:
I'm going to close this as it's no longer relevant. We have deprecated and removed the ActiveRecord persistence pattern in version 6.0 of the elasticsearch-persistence gem.

Please consider migrating your app to use the [repository pattern](https://github.com/elastic/elasticsearch-rails/tree/master/elasticsearch-persistence#the-repository-pattern) instead.

PR #660, opened 2017-01-21
Kaminari::Hooks init call is no longer needed

twe4ked:
See: https://github.com/kaminari/kaminari/commit/653143bbd885062a6bc2f3818616e2bbef88444c

For people Googling:

```
/Users/odin/Dev/twe4ked/remember/config/initializers/elasticsearch.rb:1:in `<top (required)>': uninitialized constant Kaminari::Hooks (NameError)
        from /Users/odin/.gem/ruby/2.3.1/gems/railties-5.0.1/lib/rails/engine.rb:648:in `block in load_config_initializer'
        from /Users/odin/.gem/ruby/2.3.1/gems/activesupport-5.0.1/lib/active_support/notifications.rb:166:in `instrument'
        from /Users/odin/.gem/ruby/2.3.1/gems/railties-5.0.1/lib/rails/engine.rb:647:in `load_config_initializer'
        from /Users/odin/.gem/ruby/2.3.1/gems/railties-5.0.1/lib/rails/engine.rb:612:in `block (2 levels) in <class:Engine>'
        from /Users/odin/.gem/ruby/2.3.1/gems/railties-5.0.1/lib/rails/engine.rb:611:in `each'
        from /Users/odin/.gem/ruby/2.3.1/gems/railties-5.0.1/lib/rails/engine.rb:611:in `block in <class:Engine>'
        from /Users/odin/.gem/ruby/2.3.1/gems/railties-5.0.1/lib/rails/initializable.rb:30:in `instance_exec'
        from /Users/odin/.gem/ruby/2.3.1/gems/railties-5.0.1/lib/rails/initializable.rb:30:in `run'
        from /Users/odin/.gem/ruby/2.3.1/gems/railties-5.0.1/lib/rails/initializable.rb:55:in `block in run_initializers'
        from /Users/odin/.rubies/ruby-2.3.1/lib/ruby/2.3.0/tsort.rb:228:in `block in tsort_each'
        from /Users/odin/.rubies/ruby-2.3.1/lib/ruby/2.3.0/tsort.rb:350:in `block (2 levels) in each_strongly_connected_component'
        from /Users/odin/.rubies/ruby-2.3.1/lib/ruby/2.3.0/tsort.rb:422:in `block (2 levels) in each_strongly_connected_component_from'
        from /Users/odin/.rubies/ruby-2.3.1/lib/ruby/2.3.0/tsort.rb:431:in `each_strongly_connected_component_from'
        from /Users/odin/.rubies/ruby-2.3.1/lib/ruby/2.3.0/tsort.rb:421:in `block in each_strongly_connected_component_from'
        from /Users/odin/.gem/ruby/2.3.1/gems/railties-5.0.1/lib/rails/initializable.rb:44:in `each'
        from /Users/odin/.gem/ruby/2.3.1/gems/railties-5.0.1/lib/rails/initializable.rb:44:in `tsort_each_child'

# ... snip
```

### DIFF
--- a/elasticsearch-model/README.md
+++ b/elasticsearch-model/README.md
@@ -300,10 +300,9 @@ In a Rails controller, use the the `params[:page]` parameter to paginate through
 @articles.next_page
 # => 3
 ```
-To initialize and include the Kaminari pagination support manually:
+To include the Kaminari pagination support manually:
 
 ```ruby
-Kaminari::Hooks.init
 Elasticsearch::Model::Response::Response.__send__ :include, Elasticsearch::Model::Response::Pagination::Kaminari
 ```
 


karmi:
Nice, thanks! Can you also please check [the example](https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-model/examples/activerecord_article.rb#L29), verify that it works, and remove the line also there? I'm wondering how to do this so we don't break old Kaminari versions, maybe call the method if the `Hooks` constant is defined?

twe4ked:
I guess this is going to be more of a documentation thing. Perhaps the wording should just change to mention that after version `x.x.x` you don't need the init line. We could also add a comment above the `init` call in the example.

Not sure the of the fastest way to get that example going. I can confirm it's not needed in my app. As you can see in the current version of Kaminari it's a no op.

twe4ked:
Thanks @karmi!

karmi:
Hi @twe4ked, sorry, meant to leave a note here, and forgot! Thank you for reporting the issue -- I think it's safer to go with the `if defined? ...` approach in code and documentation as well, over time we can safely remove it altogether. Thanks again!

PR #668, opened 2017-02-08
Update 03-expert to work with README.md

jsmestad:
Latest Rails versions do not have a README.rdoc, use README.md


jsmestad:
Signed the CLA

karmi:
Thanks!, merged! I've missed this change in latest Rails, so I've fixed the other templates as well, both the logic to insert the content and the logic to check whether a previous template has been already run.

PR #680, opened 2017-03-14
Cequel support

lexxiebelle:
Add support cassandra with cequel adapter. 
Create index from partition and cluster columns.


karmi:
Hi @lexxiebelle, thanks for the feature! I've got couple of notes here:

* Can you please rebase and squash the commits, probably just into one?, it's a one logical change
* The CLA check still fails, I've tried looking up your signature, but couldn't 
* Can you please add at least some rudimentary unit tests, you can adapt [the Mongoid example](https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-model/test/unit/adapter_mongoid_test.rb) I think
* Can you please integration tests for the feature, so there's a clear "shared code" for sorting out problems now in the future? Again, this can be modelled by the [Mongoid example](https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-model/test/integration/mongoid_basic_test.rb).
* I see you've implemented the `Callbacks` module as well, which is a noble thing to do :), but I see so many problems in practical scenarios with these, and people who try to (ab)use them instead of writing their own custom callbacks, that I almost think it's harmful for having them. What do you think about actually removing them and showing people how to do their own callbacks in the example and the integration test?
* Based on the missing `__find_in_batches` method implementation, I assume importing (`MyModel.import`) is not implemented. _That_ is actually -- as opposed to automatic callbacks -- something very beneficial to library users, because the `import` method has a nice customization with the transform lambda and such. Would you be OK with adding that?

estolfo:
@lexxiebelle Would you be able to address the comments above?

PR #683, opened 2017-03-22
results size is elasticsearch results get really total

houdl:
results total is elasticsearch results total, but results size is the really count.


houdl:
when i add 'size' to search dsl , i use kaminari all = A.search(xxx).page(params[:page]), i got all.total equal A.search(xxx).results.total, but it should equl A.search(xxx).size, because i use 'size' dsl to limit result from elasticsearch.

estolfo:
Hi @houdl I think the `#total_count` and `#total_entries` is intended to be the result set size so this change isn't necessary. Let me know if you agree.

estolfo:
I'm going to close this, as `#total_count` and `#total_entries` is intended to be the result set size.

PR #689, opened 2017-03-29
reindex on assoc updates only if already saved

Meekohi:
Only trigger a re-index if the Article has already been saved. This is important is you ever create an object with `.new` instead of `.create` and then add associations to it _before_ saving (note that rails is smart and will "wait" to do all the database work on `.save` in this case).

### DIFF
--- a/elasticsearch-model/examples/activerecord_associations.rb
+++ b/elasticsearch-model/examples/activerecord_associations.rb
@@ -119,8 +119,8 @@ end
 class Article < ActiveRecord::Base
   include Searchable
 
-  has_and_belongs_to_many :categories, after_add:    [ lambda { |a,c| a.__elasticsearch__.index_document if a.id } ],
-                                       after_remove: [ lambda { |a,c| a.__elasticsearch__.index_document if a.id } ]
+  has_and_belongs_to_many :categories, after_add:    [ lambda { |a,c| a.__elasticsearch__.index_document if a.persisted? } ],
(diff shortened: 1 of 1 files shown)


karmi:
Hi, that sounds legitimate -- although, would you have an example Ruby code to demonstrate how it breaks without the condition? I'd like to see it not only to satisfy my curiosity, but to better understand the flow there.

Meekohi:
So, I believe the following would exercise the error (based on the example in `examples/activerecord_associations.rb`)

```
category = Category.create title: 'One'
author = Author.create first_name: 'John', last_name: 'Smith', department: 'Business'

article  = Article.new title: 'First Article'
article.categories << category
```
```
article.authors << author
article.save!
```

karmi:
Right, thanks for pointing it out, I can see it in Pry now!

We definitely need to fix that example, I'd like to consider another way of expressing the intention there, however: instead of using `if a.id`, what if we used `if a.persisted?` That seems to me to convey the intention better, ie. _Save the document when the database record has been already stored_, as opposed to _Save the document when the database record has an ID_.

I can imagine scenarios where the latter approach would make sense, eg. when you're setting the IDs yourself in application code, but in the context of the example, which relies on the database-generated ID, the `persisted?` approach is more didactic?

Meekohi:
Yup makes sense -- wanting to push to elasticsearch with an id but before saving would be an odd use case, and now this example makes it pretty clear that you need consider the details of when you do/don't index the Article anyway.

PR #692, opened 2017-04-01
[MODEL] Fix import method to pass index name on refresh

### LINKED ISSUE #506: The "refresh", and "index" options of Model.import don't work well together
Hey,

tl;dr: in `Model.import`, `self.refresh_index!` is called without specifying the index to refresh, which I think is an issue if you specify the `index` option.

---

I have a rake task that goes something like...

```
new_index_name = args[:index_name]

Map.__elasticsearch__.create_index! index: new_index_name
Map.import(
    index: new_index_name,
    refresh: true,
    query: -> { includes(map_items: :item) }
)
```

If I run `rake map:reindex[map_index]`, then new `map_index` index will be created properly, and the maps will be imported in the right index, but **the wrong index will be refreshed**.

```
POST http://localhost:9200/alias_maps_development/_refresh
```

`alias_maps_development` being the actual index of my model (well, it's an alias).

Is this behavior intended? Shouldn't `refresh` refresh the index that's being passed as an option? (`lib/elasticsearch/model/importing.rb:130`)

Thanks

ganta:
Fixes #506

When specifying the index option and refresh option on import, the imported index is not refreshed because the specified index name is not used for refreshing.

### DIFF
--- a/elasticsearch-model/lib/elasticsearch/model/importing.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/importing.rb
@@ -130,7 +130,7 @@ module Elasticsearch
             errors +=  response['items'].select { |k, v| k.values.first['error'] }
           end
 
-          self.refresh_index! if refresh
+          self.refresh_index! index: target_index if refresh
 
           case return_value
             when 'errors'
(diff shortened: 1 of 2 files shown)


estolfo:
Thanks @ganta

PR #696, opened 2017-04-06
Fix basic template when elasticsearch exists

ryanhouston:
When attempting to generate a Rails application using the `01-basic.rb`
template the following error occurs when Elasticsearch is already
running.

```
apply  https://raw.github.com/elasticsearch/elasticsearch-rails/master/elasticsearch-rails/lib/rails/templates/01-basic.rb
https://raw.github.com/elasticsearch/elasticsearch-rails/master/elasticsearch-rails/lib/rails/templates/01-basic.rb:39:in `apply': uninitialized constant Rails::Generators::AppGenerator::JSON (NameError)
```

Adding `require 'json'` to the template file fixes this issue.

### DIFF
--- a/elasticsearch-rails/lib/rails/templates/01-basic.rb
+++ b/elasticsearch-rails/lib/rails/templates/01-basic.rb
@@ -22,6 +22,7 @@
 
 require 'uri'
 require 'net/http'
+require 'json'
 
 at_exit do
   pid = File.read("#{destination_root}/tmp/pids/elasticsearch.pid") rescue nil


karmi:
Hi @ryanhouston, it's weird, I would swear I have tested the templates with and without ES running many times during latest round of fixes, but I guess requiring JSON doesn't hurt anything, the error might be related to a specific Ruby version... 

One thing: the CLA check is red, can you please check that you have signed and that you've added the Git commit e-mail to your Github profile?

ryanhouston:
Thanks @karmi. Looks like I was using ruby 2.2.4p230 (2015-12-16 revision 53155) which is a bit outdated.

I have signed and confirmed the CLA using the same email address as attached to my Github profile. The Git commit e-mail is also the same as attached to my Github profile. Odd that it is still red.

estolfo:
This change is already incorporated into the repo as of https://github.com/elastic/elasticsearch-rails/commit/e1fc0e6df0b85d2eb65df1dec1459fc1da980dc8
Thanks!

PR #703, opened 2017-04-13
Persist attributes changed by callbacks or the save process

nbudin:
If there are `before_save` callbacks in a model that uses `Elasticsearch::Indexing`, and those `before_save` callbacks change attributes of the model, the resulting values might not be saved into Elasticsearch, because `update_document` will only pick up attributes that were changed by the time its own `before_save` callback ran.  Similarly, if the model indexes `updated_at`, it will never be saved into Elasticsearch because that value is updated as part of the save process itself.

This PR changes the behavior of `Elasticsearch::Indexing` to set the `@__changed_attributes` hash in an `after_save` callback, using the `previous_changes` method.  This will contain any attributes changed by other callbacks or the save process itself.

(Also, I'm working on having my company sign the CLA and will let you know when that's done.)

### DIFF
--- a/elasticsearch-model/lib/elasticsearch/model/indexing.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/indexing.rb
@@ -314,10 +314,10 @@ module Elasticsearch
           #
           # @see #update_document
           #
-          base.before_save do |instance|
+          base.after_save do |instance|
             instance.instance_variable_set(:@__changed_attributes,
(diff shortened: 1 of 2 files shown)


nbudin:
PatientsLikeMe has now signed the CLA and I've also signed an individual one under PatientsLikeMe.

nbudin:
@karmi - ping?  Sorry to bug you.

PR #715, opened 2017-05-23
Error raised with Rails 5.0.0  / Ruby 2.3.3

theoo:
The following error is raised
01-basic.rb:39:in `apply': uninitialized constant #<Class:#<Rails::Generators::AppGenerator:0x00000000e3da28>>::JSON (NameError)
without this line.


estolfo:
I'm going to close this as the change is already included.
Please see [this line](https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-rails/lib/rails/templates/01-basic.rb#L24).

PR #717, opened 2017-06-01
Fixed an infinite recursion bug in STI handling code

stevemadere:
This is a fix suggested by kayakyakr in the comments on PR #332.
It solves the stack-too-deep problem which I encountered as soon as I enabled the
inheritance support.


jsmecham:
This appears to resolve the same problem that I am having.

tmandke:
Same as #585 and that one has some tests as well.

PR #720, opened 2017-06-15
[MODEL] Fix import when preprocess returns empty collection

### LINKED ISSUE #719: `Scope#import` with `options[:preprocess]` throws exception when preprocessed data returns empty
Hej!

I have a model `person` and I want to import all `published` people. All not yet published people will be rejected from the `batch` in the `preprocess`. So it is possible, that all items will be removed from the `batch` and I am sending an empty `batch`.  That will throw the following exception:

```
Elasticsearch::Transport::Transport::Errors::BadRequest: [400] {"error":{"root_cause":[{"type":"parse_exception","reason":"Failed to derive xcontent"}],"type":"parse_exception","reason":"Failed to derive xcontent"},"status":400}
```

I think the behavior should be like `import` empty scoped data.

timbleck:
Closes #719 and also fixes existing test on preprocess which had no effect.

### DIFF
--- a/elasticsearch-model/lib/elasticsearch/model/adapters/active_record.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/adapters/active_record.rb
@@ -103,7 +103,8 @@ module Elasticsearch
             scope = scope.instance_exec(&query) if query
 
             scope.find_in_batches(options) do |batch|
-              yield (preprocess ? self.__send__(preprocess, batch) : batch)
+              batch = self.__send__(preprocess, batch) if preprocess
+              yield(batch) if batch.present?
             end
           end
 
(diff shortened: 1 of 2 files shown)


aeroastro:
Great work!
I have been using 5.x version on production environment with a similar patch.
I'm wondering when this will be merged into all the `elasticsearch-rails` versions.

karmi:
@timbleck , @aeroastro , sorry for the long silence here! As I've mentioned on other tickets, we're scheduling a significant effort to give attention to the `elassticsearch-rails` groups of libraries again in couple of weeks, so I'd rather not merge the PR now, but wait a bit, until we have more capacity and focus to be more organized with that...

estolfo:
thanks @timbleck !

PR #732, opened 2017-07-17
[MODEL] Added the `:includes` option to Adapter::Multiple::Records

mbv:
This patch adds the `:includes` option to Multiple adapter
which can be used to eager load submodels just like
you'd do with regular ActiveRecord.

This helps to eliminate the N+1 problem when using
elasticsearch-model searches.

Example:

    class Article < ActiveRecord::Base
      belongs_to :author

      # ...

    end

    class Comment < ActiveRecord::Base
      belongs_to :article
      belongs_to :user

      # ...

    end

    Elasticsearch::Model.search(..., [Article, Comment]).records(includes: {Article: [:author], Comment: [:article, :user]})

    # Would be similar to using:

    Article.includes(:author).where(...)
    Comment.includes(:article, :user).where(...)

### DIFF
--- a/elasticsearch-model/lib/elasticsearch/model/adapters/multiple.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/adapters/multiple.rb
@@ -11,6 +11,12 @@ module Elasticsearch
         Adapter.register self, lambda { |klass| klass.is_a? Multimodel }
 
         module Records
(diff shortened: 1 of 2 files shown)


pocke:
Any news?
I'd like to avoid N+1, so I'm happy if this pull request is merged.
It looks good to me.

Issue #313, opened 2015-01-08
Error when using basic template


bgrayburn:
Running:
```
rails new searchapp --skip --skip-bundle --template https://raw.github.com/elasticsearch/elasticsearch-rails/master/elasticsearch-rails/lib/rails/templates/01-basic.rb
```
Returning:
```       
apply  https://raw.github.com/elasticsearch/elasticsearch-rails/master/elasticsearch-rails/lib/rails/templates/01-basic.rb
/Users/bgrayburn/.rvm/rubies/ruby-2.0.0-p247/lib/ruby/2.0.0/net/http.rb:918:in `connect': SSL_connect returned=1 errno=0 state=SSLv3 read server key exchange B: bad ecpoint (OpenSSL::SSL::SSLError)
```

karmi:
This runs fine for me -- it might have been a problem on the Github side. Please try it again and post an update!

bgrayburn:
Thanks karmi, tried it again and it worked up until I got this:
```
     gemfile    mocha
/media/Storage/firstmilegeo/elasticSearchTest/01-basic.rb:123:in `apply': undefined method `comment_lines' for #<Rails::Generators::AppGenerator:0x007fa93982ca78> (NoMethodError)
	from /Users/bgrayburn/.rvm/gems/ruby-2.0.0-p247/gems/thor-0.14.6/lib/thor/actions.rb:217:in `instance_eval'
	from /Users/bgrayburn/.rvm/gems/ruby-2.0.0-p247/gems/thor-0.14.6/lib/thor/actions.rb:217:in `apply'
	from /Users/bgrayburn/.rvm/gems/ruby-2.0.0-p247/gems/railties-3.2.0/lib/rails/generators/app_base.rb:109:in `apply_rails_template'
	from (eval):1:in `apply_rails_template'
	from /Users/bgrayburn/.rvm/gems/ruby-2.0.0-p247/gems/thor-0.14.6/lib/thor/task.rb:22:in `run'
	from /Users/bgrayburn/.rvm/gems/ruby-2.0.0-p247/gems/thor-0.14.6/lib/thor/invocation.rb:118:in `invoke_task'
	from /Users/bgrayburn/.rvm/gems/ruby-2.0.0-p247/gems/thor-0.14.6/lib/thor/invocation.rb:124:in `block in invoke_all'
[...]

PR #728, opened 2017-07-07
Update 01-basic.rb

vitor-caetano:
added the required json file.

PR #733, opened 2017-08-01
required json for end of issue #313

tostasqb:


psylone:
Same as #728 PR.

psylone:
Since this change doesn't affect the test suit, can it be merged?

nilbus:
BTW, as harmless as this PR would be, I couldn't reproduce this problem with rails 3.2, 4.2, 5.0, or 5.1 on MRI and JRuby. I imagine `rails new` is requiring the json library before this template script is executed. Clearly others are experiencing this problem though, and I wonder if some combination of dependencies is causing it.

tostasqb:
Signed the licence agreement

estolfo:
I can't reproduce this but if there's an error for someone, I don't see a problem with requiring `json`.

PR #747, opened 2017-10-09
add travis badge to README

lostapathy:


lostapathy:
I missed the badge mid-page, suggest we move them to the top as that seems to be the convention.

estolfo:
thanks @lostapathy

PR #748, opened 2017-10-09
update download URL and ruby versions for travis

lostapathy:
Other builds are failing due to travis pointing to a download snapshot that no longer exists.  Updated that and ruby versions to attempt to fix that.


lostapathy:
Closing this, working on a bigger PR

PR #750, opened 2017-10-12
Avoid overriding previously defined custom options for lograge

bolshakov:
Prevent overriding `lograge.custom_options` defined in main app.

### DIFF
--- a/elasticsearch-rails/lib/elasticsearch/rails/lograge.rb
+++ b/elasticsearch-rails/lib/elasticsearch/rails/lograge.rb
@@ -29,8 +29,18 @@ module Elasticsearch
             include Elasticsearch::Rails::Instrumentation::ControllerRuntime
           end
 
+          previous_lograge_custom_options = config.lograge.custom_options
           config.lograge.custom_options = lambda do |event|
-            { es: event.payload[:elasticsearch_runtime].to_f.round(2) }
+            previous_lograge_custom_options_hash = if previous_lograge_custom_options
+              if previous_lograge_custom_options.respond_to?(:call)
+                previous_lograge_custom_options.call(event)
+              else
+                previous_lograge_custom_options
+              end
+            else
+              {}
+            end
+            previous_lograge_custom_options_hash.merge(es: event.payload[:elasticsearch_runtime].to_f.round(2))
           end
         end
       end


halilim:
One thing to note: [`custom_options ` can be a `Hash`](https://github.com/roidrage/lograge/blob/822cc26f9e5353a4b52bd148b7b6775d5719595c/lib/lograge.rb#L22).

I think a better option would be Lograge supporting an array/custom object that supports appending/prepending to it.

elasticsearch-rails/lib/elasticsearch/rails/lograge.rb
```diff
@@ -33,8 +33,14 @@ class Railtie < ::Rails::Railtie
             include Elasticsearch::Rails::Instrumentation::ControllerRuntime
           end
 
+          previous_lograge_custom_options = config.lograge.custom_options
           config.lograge.custom_options = lambda do |event|
-            { es: event.payload[:elasticsearch_runtime].to_f.round(2) }
+            previous_lograge_custom_options_hash = if previous_lograge_custom_options
+              previous_lograge_custom_options.call(event)
```
nilbus: Good catch @halilim. This assumes that `previous_lograge_custom_options` responds to `call`, which it won't when it's a Hash.

bolshakov:
I'm absolutely agree that having multiple `custom_options` is more flexible. Currently I have the following fix in my `application.rb` and it works an applications side

```ruby
  class Application < Rails::Application
    # Load it after ELC's lograge engine to fix overriding custom_options
    # @see https://github.com/elastic/elasticsearch-rails/issues/749
    initializer 'custom_lograge', after: 'elasticsearch.lograge' do
      previous_lograge_custom_options = config.lograge.custom_options
      config.lograge.custom_options = lambda do |event|
        previous_lograge_custom_options.call(event).merge(
          user: event.payload[:user],
        )
      end
    end
  end
```

I updated this PR as well to support hash in `custom_options`.

leiz-me:
When will this PR get into the master or any release version that may update with bundler?

estolfo:
@bolshakov can you try signing the CLA again please? Thanks

bolshakov:
Unfortunately it still says I have disabled cookies.

estolfo:
Would you mind trying in a different browser?

bolshakov:
I managed to open the agreement and unfortunately can't assess this legal document and possible consequences for me. So, I'm closing this PR.

PR #754, opened 2017-11-08
Fix travis, simplify Rakefile, improve coverage

lostapathy:
Apologies that this is a big PR, but not sure how to split this up without feedback and with Travis broken.  I am happy to split this into smaller PRs, or adjust in general, based on feedback.

Several things have gone on here:

1)  Update the download URLs for travis so that the project builds
2)  Support for rails versions < 4.2 have been removed since they are no longer supported upstream
3)  Added the Appraisal gem to manage gemfiles and did some initial work to simplify the Rakefiles based on that.
4)  Updated gemfiles and removed obsolete gemfiles based on prior points
5)  Removed some ruby-1.x code since we no longer support that
6)  Commented out 3 tests in elasticsearch-model/test/integration/active_record_associations_parent_child_test.rb that I believe are obsolete due to https://www.elastic.co/guide/en/elasticsearch/reference/6.0/breaking-changes-6.0.html


lostapathy:
dd107c0 needs to be backed out after a new version of elasticsearch-extensions is cut.

lostapathy:
If/when this gets upstream there's more we can do to streamline the travis config, but I'll wait for that in a separate PR

lostapathy:
Latest commits remove the Gemfile changes that made this not ready to merge - this is ready to go now.  thanks!

lostapathy:
> FYI, I opened #763 to use 6.1.0, which was released 2 days ago.

Interestingly, elastic/elasticsearch-ruby is now testing against 7.0-alphas, FWIW.

> Was the problem you hit fixed in elasticsearch-extensions 0.0.27 released in November? (CHANGELOG)

Yes, and my 588a45e undid that.


> This does make the build green, but I would hesitate to merge commented tests that identify seemingly real failures.

They are real failures, but near as I can tell they are testing things  Elasticsearch 6 dropped support for, see the changelog linked above.

What really needs to happen here, IMO, is this build (and elasticsearch-ruby) need to be further refactored to support a build matrix that would test the gem against ES5, 6, and 7.  With failures allowed against 7 (it's alpha, after all) and tests updated to not run against ES versions the functionality doesn't work with.  We also need to split this up to have travis build each gem separately, rather than as a group, to aid in troubleshooting.  With that in hand, it'd be a lot easier to get tests un-skipped and really wring things out well.

But ... I don't have the energy to work any more of this out until we start getting some engagement both here and on elastic/elasticsearch-ruby to merge the early phases of such cleanup.  To go much farther will create a PR that's *really* hard to feel good about merging.

nilbus:
Yes, that all makes sense to me.

gfx:
+1

estolfo:
I'm going to close this PR because I don't believe it's relevant anymore.

Issue #161, opened 2014-07-07
Support "wrapper" option


benjaminbarbe:
Wrapping with Hashie::Mash can be slow (~ 4s)
![with-hashie](https://cloud.githubusercontent.com/assets/630365/3496217/e38a0c24-05df-11e4-96f8-6ffb5d098b5e.png)

Versus [extend method](https://gist.github.com/benjaminbarbe/99be9a4391de1a1284b0) (but it won't allow deeper hierarchies ~ 0.26s)
![without-hashie](https://cloud.githubusercontent.com/assets/630365/3496218/e38af788-05df-11e4-928b-f1015fee9bb0.png)

One option would be to support "wrapper" option, which could be Hash, Hashie::Mash, custom object...
`Elasticsearch::Model.wrapper = Hash`

fbernier:
This would indeed be a useful feature. Would you still accept a PR for this?

Animeye:
Has there been any further information on this?

karmi:
@Animeye No new information, this felt off the radar... If anybody would like to pick it up (@fbernier), I'm up for assistance and review!

martinemde:
@karmi I submitted a partial fix for this in #755. Hopefully it can address some of the concerns in this thread.

bkazez:
I'm still seeing large performance issues due to `Hashie::Mash` - see below. What is the recommended workaround?

<img width="800" alt="screen shot 2017-12-20 at 01 06 49" src="https://user-images.githubusercontent.com/64005/34199285-262c9cca-e522-11e7-98d5-1669712f02e9.png">

martinemde:
@bkazez I monkeypatched the gem for now to remove the wrapping in a Hashie::Mash. It has made a big difference for me. I also submitted #755 to attempt to deal with it without breaking anything.

btw, here's the monkeypatch we are using in our production app:

```ruby
MonkeyPatch.for('elasticsearch-persistence', '<your gem version here>') do
  module Elasticsearch
    module Persistence
      module Repository
        module Response # :nodoc:
          class Results
            # This patch removes the usage of Hashie::Mash which is very very slow
            # and provides us no value. As of 2017-11-15, the newest version of
            # elasticsearch-persistence gem still has a compatible initialize
            # method that won't be broken by this patch.
            #
            # https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-persistence/lib/elasticsearch/persistence/repository/response/results.rb
            #
[...]

PR #755, opened 2017-11-16
Avoid unnecessarily incurring the cost of HashWrapper results

martinemde:
The HashWrapper is very expensive to create. It goes unused unless
specifically accessed. Without breaking backwards compatibility,
this change allows the HashWrapped response to be available, but not
computed for every result by default. This significantly improves
the performance of search results in my tests.

This addresses part of the performance concern in #161 while maintaining backwards compatibility. If you only use the deserialized results, you'll never need to create the HashWrapper. Only if you actually use the response object will you get the wrapped results.

In our app, the creation of a Hashie::Mash was responsible for about 300ms of our 770ms response time. This action only ever used the deserialized result (which is sourced by converting to a Mash, and then back to a Hash) and never the response. Patching this to remove the Mash entirely makes a huge difference in performance.

I hope this can silently help others so they don't need to dig into this particular performance issue.

### DIFF
--- a/elasticsearch-persistence/test/unit/repository_response_results_test.rb
+++ b/elasticsearch-persistence/test/unit/repository_response_results_test.rb
@@ -45,6 +45,14 @@ class Elasticsearch::Persistence::RepositoryResponseResultsTest < Test::Unit::Te
       assert_equal 5, subject.response._shards.total
     end
 
+    should "provide access to the raw response" do
+      assert_equal 5, subject.raw_response['_shards']['total']
+    end
+
+    should "return the raw response unwrapped" do
+      assert_equal @response, subject.raw_response
+    end
+
     should "return the total" do
       assert_equal 2, subject.total
     end
--- a/elasticsearch-persistence/lib/elasticsearch/persistence/repository/response/results.rb
+++ b/elasticsearch-persistence/lib/elasticsearch/persistence/repository/response/results.rb
@@ -18,7 +18,7 @@ module Elasticsearch
           #
           def initialize(repository, response, options={})
             @repository = repository
-            @response   = Elasticsearch::Model::HashWrapper.new(response)
+            @response   = response
             @options    = options
           end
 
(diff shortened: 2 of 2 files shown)


martinemde:
Please let me know what I can do to fix the checks. I signed the CLA, but I don't see what I can do for CI.

karmi:
Hi @martinemde, this looks promising, thanks!, I understand that creating `Hashie::Mash` is expensive.

Regarding CLA, I can't find you in our records by name -- can you ping me by email, karel@elastic.co, so we can sort it out?

CI is problematic for some time, I'll run the tests locally.

I can't guarantee I'll process the patch in terms of days, but I'll try to look into it. Please ping me again by mentioning me on the ticket if I forget about it. Thanks!

martinemde:
@karmi I updated a bit since I found this exact patch would not solve all our issues. We'd still end up creating the Mash even though we just access the result one time with `results["aggregations"]`.

Looks like the CLA glitched. I tried again and it went through to my email this time. I guess I missed something the first time.

elasticsearch-persistence/test/unit/repository_response_results_test.rb
```diff
@@ -45,6 +45,14 @@ class MyDocument; end
       assert_equal 5, subject.response._shards.total
     end
 
+    should "provide access to the raw respanse" do
```
guillaumebriday: There is a typo with `respanse` instead of `response` right ?

martinemde: @guillaumebriday Thanks for noticing. Fixed.

estolfo:
@martinemde I've pinged you [here](https://github.com/elastic/elasticsearch-rails/issues/161#issuecomment-411766031). Let me know if the pull request helps with the performance.

estolfo:
Closing this, please see my comment here https://github.com/elastic/elasticsearch-rails/issues/161#issuecomment-418713736

PR #758, opened 2017-11-29
Fix deprecation warning from Rails 5.1 upgrade (changed_attributes is now attributes_in_database)

Geesu:
This PR only changes 2 things to fix the deprecation warning:
1. `changed_attributes` changes to `attributes_in_database`
2. `instance.changes` changes to `instance. changes_to_save`

### DIFF
--- a/elasticsearch-model/lib/elasticsearch/model/indexing.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/indexing.rb
@@ -307,7 +307,7 @@ module Elasticsearch
 
         def self.included(base)
           # Register callback for storing changed attributes for models
-          # which implement `before_save` and `changed_attributes` methods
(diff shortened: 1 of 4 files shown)


karmi:
Hi @Geesu, first of all, many apologies for the delay with processing this patch...

I've merged your patch, and added another one, which restores the functionality for older Rails versions (since I don't want to break the compatibility in an isolated commit, without proper deprecations and version bumps). I've also changed the conditions in 7815039 to use the `changes_to_save` / `changes` method names to clearly indicate the intent: my original code using the `changed_attributes` was very confusing.

Can you please evaluate this version? I've also pushed it to the `5.x` branch, and released it as [`5.0.2`](https://rubygems.org/gems/elasticsearch-model/versions/5.0.2) on Rubygems.

PR #762, opened 2017-12-15
[MODEL] Avoid making an update when no attributes are changed

### LINKED ISSUE #743: Sometime `@__changed_attributes` have {} value.
```
elasticsearch-rails (5.0.1)
rails (= 5.1.3)
```

ActiveRecord's `update_attributes` does return `{}` from `instance.changes` when don't have changed record.

`update_attribute` runs below.

``` ruby
pry(main)> job = Job.first
pry(main)> puts job.title
Updated1
pry(main)> job.update_attribute(:title, 'Updated1')
pry(main)> puts job.__elasticsearch__.instance_variable_get(:@__changed_attributes)

=> nil
```

`update_attributes` runs below.

``` ruby
pry(main)> job = Job.first
pry(main)> puts job.title
Updated1
pry(main)> job.update_attributes(title: 'Updated1')
pry(main)> puts job.__elasticsearch__.instance_variable_get(:@__changed_attributes)
{} # what's up!
=> nil
```

This behavior affects `def update_document(options={})` method.

https://github.com/elastic/elasticsearch-rails/blob/5.x/elasticsearch-model/lib/elasticsearch/model/indexing.rb#L394

`if changed_attributes = self.instance_variable_get(:@__changed_attributes)` evaluated `true` but `changed_attributes` is `{}`.
Finally, send request to elasticsearch empty update.

I guess `@__changed_attributes` should not have `{}`, It is correct specification this gem?

nilbus:
@dim0627 pointed out in #743 that calling `update_attributes({})` on an ActiveRecord model instance triggers the `before_save` callback, even when there are no attributes to update. This results in an unnecessary empty update request to elasticsearch.

Using ActiveSupport's `presence`, it will act the same for both `nil` and `{}`, which are both ways Rails can express that there are no changes to save.

Closes #743

### DIFF
--- a/elasticsearch-model/lib/elasticsearch/model/indexing.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/indexing.rb
@@ -397,7 +397,7 @@ module Elasticsearch
         # @see http://rubydoc.info/gems/elasticsearch-api/Elasticsearch/API/Actions:update
         #
         def update_document(options={})
-          if attributes_in_database = self.instance_variable_get(:@__changed_model_attributes)
+          if attributes_in_database = self.instance_variable_get(:@__changed_model_attributes).presence
             attributes = if respond_to?(:as_indexed_json)
               self.as_indexed_json.select { |k,v| attributes_in_database.keys.map(&:to_s).include? k.to_s }
             else
--- a/elasticsearch-model/test/unit/indexing_test.rb
+++ b/elasticsearch-model/test/unit/indexing_test.rb
@@ -176,6 +176,19 @@ class Elasticsearch::Model::IndexingTest < Test::Unit::TestCase
         end
       end
 
+      class ::DummyIndexingModelWithNoChanges
+        extend  Elasticsearch::Model::Indexing::ClassMethods
+        include Elasticsearch::Model::Indexing::InstanceMethods
+
+        def self.before_save(&block)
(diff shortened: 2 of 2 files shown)


dim0627:
Thank you for making PR for my problem though 🙏 

I think this changes should use `presence` instead of `.present?`.
Or `if (attributes_in_database = self.instance_variable_get(:@__changed_model_attributes)).present?`.

Otherwise, boolean will be assigned to `attributes_in_database`.

nilbus:
Fantastic catch. I had interpreted it as a boolean, but the tests were just telling me the same thing. 😄

nilbus:
Tests updated and passing (locally).

nilbus:
@dim0627 I should point out that this change only replaces the `update` call with an `index` call.

https://github.com/elastic/elasticsearch-rails/blob/d8859aed2aa75c93840b763d70e8fce8ce7761ff/elasticsearch-model/lib/elasticsearch/model/indexing.rb#L407-L409

This change makes the behavior more consistent, but is it actually going to be helpful to you?

dim0627:
@nilbus 
Yes, I confused `Why don't called index_document!?`.
Your change is looks good to me 👍

estolfo:
@nilbus and @dim0627 thanks for your work on this! I've tested the changes locally and will merge.

nilbus:
Thanks! :tada:

Issue #756, opened 2017-11-16
ES 6.x support


peter-hank:
Does the gem support Elasticsearch 6.x? If not, do you plan to support it anytime soon?

karmi:
Hi @peter-hank, the libraries should _in general_ support Elasticsearch 6.x, but it might require some additional configuration, namely when it comes to [removal of types](https://www.elastic.co/guide/en/elasticsearch/reference/6.x/removal-of-types.html). Eg. the `elasticsearch-model` in the default configuration will try to use multiple types per model class, so a user configuration which makes it use a separate index for each model class might be needed.

Also, the tests and examples haven't been yet updated to 6.x, which is something needed... We do plan to support 6.x as soon as possible in the default configuration of the gems, but I don't have a concrete date.

sly7-7:
@karmi In order to test this, I guess I must use the master branch ? (Also, maybe you want some help to update the tests and examples ?)

Aesthetikx:
Just anecdotal evidence, I have been seeing HTTP 406 errors on 6.x, haven't dug into why though.

```
Failure/Error: model.__elasticsearch__.create_index!(force: true)
     
     Elasticsearch::Transport::Transport::Errors::NotAcceptable:
       [406] {"error":"Content-Type header [application/x-www-form-urlencoded] is not supported","status":406}
```

tenaz3:
How can I have the elasticsearch 6 with elasticsearch-rails 5 ?

```Bundler could not find compatible versions for gem "elasticsearch":
  In Gemfile:
    elasticsearch (~> 6.0)

    elasticsearch-model (>= 5.0.2, ~> 5.0) was resolved to 5.0.2, which depends on
      elasticsearch (~> 5)

    elasticsearch-model (>= 5.0.2, ~> 5.0) was resolved to 5.0.2, which depends on
      elasticsearch (~> 5)
```

karmi:
@tenaz3, so what is exactly the problem with 6 and 5? The `elasticsearch-model` has a [very relaxed dependency](https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-model/elasticsearch-model.gemspec#L26) on the Elasticsearch client. I've just generated a new Rails 5, and when added the following to the Gemfile:

```ruby
gem 'elasticsearch', '~> 6'
gem 'elasticsearch-model', '~> 5'
```

this is what I see after running `bundle install`:

```
$ grep 'elasticsearch' Gemfile.lock 
[...]

PR #763, opened 2017-12-15
Use Elasticsearch 6.1.0 for Travis builds

nilbus:
This allows the tests to run again, though a few are failing as mentioned in https://github.com/elastic/elasticsearch-rails/issues/756#issuecomment-344932630.

e.g.

> Rejecting mapping update to [questions_and_answers] as the final mapping would have more than 1 type: [question, answer]

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,14 +30,14 @@ matrix:
 
     - rvm: 2.3.3
       jdk: oraclejdk8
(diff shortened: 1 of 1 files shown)


leifcr:
Added #797 PR which also cleans out old Rails and Ruby versions

estolfo:
I'm going to close this, as we have the tests running against Elasticsearch 6 in the [6.x branch](https://github.com/elastic/elasticsearch-rails/tree/6.x).

PR #769, opened 2018-01-17
Extract lograge module dependencies, for merge it with existing settings

robertomiranda:
Currently is not possible merge the elastic lograge options with exiting option modules. With this change the custom options will look like

```ruby
Rails.application.configure do
  config.lograge.enabled = true
  config.lograge.custom_options = lambda do |event|
    require 'elasticsearch-rails/lib/elasticsearch/rails/lograge_modules'
    { :name => "value", 
      :timing => some_float.round(2), 
      :host => event.payload[:host], 
      :es => event.payload[:elasticsearch_runtime].to_f.round(2} }
  end
end
```

### DIFF
--- a/elasticsearch-rails/lib/elasticsearch/rails/lograge.rb
+++ b/elasticsearch-rails/lib/elasticsearch/rails/lograge.rb
@@ -17,21 +17,7 @@ module Elasticsearch
       #
       class Railtie < ::Rails::Railtie
         initializer "elasticsearch.lograge" do |app|
-          require 'elasticsearch/rails/instrumentation/publishers'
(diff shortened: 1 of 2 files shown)


nilbus:
@robertomiranda This PR addresses the same problem as #750. It looks like your approach would work. The thing I like about #750 is that users can set `config.lograge.custom_options` without having to know to do this special `require` first. Also, if the `:es` value set by this gem changes, the user won't accidentally override the new value with the old one hard-coded in the `config.lograge.custom_options` in the application, as would happen in your example.

Thanks for taking the time to put together this PR. It shows that we need to get some solution merged for custom lograge options.

PR #771, opened 2018-01-24
Updated README.MD

anandvc:
Include the step to create the index for the model that we'll import.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,7 @@ class Article < ActiveRecord::Base
   include Elasticsearch::Model::Callbacks
 end
 
+Article.__elasticsearch__.create_index!
 Article.import
 
 @articles = Article.search('foobar').records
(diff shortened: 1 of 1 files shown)


estolfo:
Let's keep the example simple and not include this for now.

anandvc:
@estolfo Does the example work for you without that line which creates the index?

estolfo:
I was hoping to keep the example as simple as possible and index creation doesn't always happen at the same time as import jobs. Is it really unclear without the `create_index!` line?

anandvc:
@estolfo Yes. The Readme does not mention index creation at all, but it is necessary for the example to work. If you want, we can add a comment for the create_index! line to say that this can be done asynchronously.

estolfo:
@anandvc I think a comment would be better, so index creation right at import time wouldn't be encouraged. I think that operation would usually be done at another point.
I'll reopen the issue if you want to update the PR?

anandvc:
Thanks, @estolfo! I've updated the PR and added a comment.

estolfo:
thanks!

PR #780, opened 2018-03-27
Added condition for :update after commit & rescue if record NotFound

romkaspb:
Hey guys
I just added readme condition for such case:
Instance of article exists, persisted, and published so successfully indexed. Then, if you make it unpublished, according to current example it will not be indexed, because 
```ruby
self.published? == true
```
I've added condition to delete document if it's non-published, and update if published. In case document was published (created),  then deleted (on update hook), and then published again, `__elasticsearch__.update_document` will create a new document.
Second point is exception handling - when gem trying to delete non-existing document, it throws `Elasticsearch::Transport::Transport::Errors::NotFound`. I've just added `rescue` for it.

Thank you.

### DIFF
--- a/elasticsearch-model/README.md
+++ b/elasticsearch-model/README.md
@@ -536,7 +536,11 @@ class Indexer
         record = Article.find(record_id)
         Client.index  index: 'articles', type: 'article', id: record.id, body: record.__elasticsearch__.as_indexed_json
       when /delete/
-        Client.delete index: 'articles', type: 'article', id: record_id
+        begin
+          Client.delete index: 'articles', type: 'article', id: record_id
+        rescue Elasticsearch::Transport::Transport::Errors::NotFound
+          logger.debug "Article not found, ID: #{record_id}"
+        end
       else raise ArgumentError, "Unknown operation '#{operation}'"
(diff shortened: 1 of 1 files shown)


karmi:
Hi @romkaspb , sorry for the long silence here! I've read through the patch and both additions do make sense, thanks! Particularly the check for `NotFound` is useful for somebody copy&pasting code :)

The CLA check is red, would be so kind to sign the [contributor agreement](http://www.elasticsearch.org/contributor-agreement/)? I understand this is a small change to documentation, but it just makes all the administration around code changes easier.

Also, if you'd have a bit more time, would you be able to split the commit into two? These are unrelated changes and I had to squint for a while until I have realized the patch touches _two_ different parts of the codebase :) It's not required by any means, I can do that myself during merging, or keep it as it is.

romkaspb:
Hi @karmi!
Sorry for much more long silence. I think I've missed it somewhere in my mailbox. :)
Just signed contributor agreement and splitter PR to different commits.
Sorry for my poor english.

Thank you!

estolfo:
Thanks @romkaspb !

PR #785, opened 2018-04-06
Support :scope, :query and :preprocess importing options for Mongoid

Seitk:
The document said It's possible to import only records from a specific scope or query and change by a preprocess function but it's not implemented for the Mongoid Adapter


Seitk:
Can't build with latest elasticsearch (elasticsearch-6.0.0-alpha1-SNAPSHOT.tar.gz), i will create a PR into 5.x first

PR #786, opened 2018-04-06
[MODEL] Support :scope, :query and :preprocess importing options for Mongoid

Seitk:
The elasticsearch model readme said It's possible to import only records from scope or query and preprocess batch data but it's not implemented

### DIFF
--- a/elasticsearch-model/lib/elasticsearch/model/adapters/mongoid.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/adapters/mongoid.rb
@@ -63,10 +63,17 @@ module Elasticsearch
           # @see https://github.com/karmi/retire/pull/724
           #
           def __find_in_batches(options={}, &block)
-            options[:batch_size] ||= 1_000
+            batch_size = options[:batch_size] || 1_000
+            query = options[:query]
+            named_scope = options[:scope]
+            preprocess = options[:preprocess]
+
+            scope = all
+            scope = scope.send(named_scope) if named_scope
+            scope = query.is_a?(Proc) ? scope.class_exec(&query) : scope.where(query) if query
   
-            all.no_timeout.each_slice(options[:batch_size]) do |items|
-              yield items
+            scope.no_timeout.each_slice(batch_size) do |items|
+              yield (preprocess ? self.__send__(preprocess, items) : items)
             end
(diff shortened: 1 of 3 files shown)


karmi:
@Seitk , this sounds right! Since we're planning some re-focus to the libraries soon, I'd rather not merge it now, in a couple of weeks we should have more capacity to give your PR and other PRs more attention...

elasticsearch-model/lib/elasticsearch/model/adapters/mongoid.rb
```diff
@@ -64,9 +64,16 @@ module Importing
           #
           def __find_in_batches(options={}, &block)
             options[:batch_size] ||= 1_000
+            query = options.delete(:query)
+            named_scope = options.delete(:scope)
+            preprocess = options.delete(:preprocess)
```
estolfo: Do you mind if we just set the `:query`, `:scope` and `:preprocess` options instead of deleting them from the `options` hash? I'd rather not mutate the `options` that the user passes to the `#import` method. Thanks!

estolfo:
Hi @Seitk I added a comment, let me know what you think so we can get this merged. Thanks!

Seitk:
@estolfo updated, should i change base on this PR or open a new one lol?

estolfo:
@Seitk Sure you can update this PR, thanks!

estolfo:
@Seitk Also, perhaps you want to follow what I've done [here](https://github.com/elastic/elasticsearch-rails/pull/827) to use the `default_scope` on the model? It seems that we should be consistent between the Mongoid and ActiveRecord adapters.

Seitk:
@estolfo The default scope is supported by default with the "all" criteria, added integration test for mongoid with default scope anyway

Seitk:
btw I am using "5.x" for a production service, i will create another PR for "6.x" separately

estolfo:
@Seitk Thanks for the reminder about the default_scope via `all`. I've updated my pull request - it turns out that it's working as expected without code changes. 
In any case, thanks for adding the test for default_scope in this PR.
The changes for the 6.x should be identical, I don't think any of this code changed in the 6.x branch so I look forward to receiving another PR. Thanks!

PR #809, opened 2018-06-29
[STORE] Warn that the ActiveRecord persistence pattern is deprecated

estolfo:


holidayworking:
Why the ActiveRecord Pattern is deprecated?

PR #814, opened 2018-07-18
[MODEL] Use default doc type: _doc

estolfo:


sonalkr132:
This bit me quite ugly :sweat_smile: If migrating from existing ES 5 cluster, make sure you set `document_type` to `self.model_name.element` or else you will end up creating new type `_doc` which won't have any documents. 

changelogs does mention:
> Use default doc type: _doc (#814)

perhaps it would be helpful if we mark this breaking change.

PR #824, opened 2018-08-06
[STORE] Refactor Repository as mixin

estolfo:
This pull request offers an alternative refactor to [this pull request](https://github.com/elastic/elasticsearch-rails/pull/822) of `Elasticsearch::Persistence::Repository`. It provides a mixin that users can include to get Repository behavior on classes.

example:
```ruby
class MyTestRepository
  include Elasticsearch::Persistence::Repository
  client DEFAULT_CLIENT
end

repo_instance = MyTestRepository.new(index_name: 'my_test_repository', document_type: 'test')
other_repo_instance = repo_instance.with(index_name: 'other_test_repository')
repo_instance.index_name != other_repo_instance.index_name
repo_instance.document_type == other_repo_instance.document_type
repo_instance.client == other_repo_instance.client
repo_instance.client == MyTestRepository.client
```


jits:
(Following on from the discussion on https://github.com/elastic/elasticsearch-rails/pull/822)

This looks great! :) Appreciate you going to the effort to put this together.

Some questions / thoughts:

1. Given you can pass in `options` to each new instance, to configure just that instance, do you still need the class level settings (like `index_name`)? I can see how they provide added flexibility (to configure defaults at the class level), so this isn't a big deal (in my opinion), but just something that consumers need to be aware of (especially as you also have defaults for these class level options, that get used if a value is not provided at all by the consumer, instead of erroring).
1. Is there much need for the `.with` helper? Don't get me wrong, it does seem like a useful helper, but given that it does a shallow `.clone` it's not certain that you'll get a completely isolated instance. Instead I'd advocate for consumers creating a fresh new instance when they need and managing common settings themselves (as with regular classes and objects). 
1. Alternatively to the above point, if a helper is really needed to instantiate a similar repo, you could have a class level helper method `MyRepo.from(my_repo_instance, new_options)` which just returns `MyRepo.new(my_repo_instance.options.merge(new_options))`?
1. This covers nicely the need for flexibility in creating different – and isolated – repository instances that could talk to different clusters and/or indexes (with the caveat of the comment above); I'd still suggest having a `GenericRepository` class that can be used to instantiate a very basic repository without having to extend anything. This is certainly not crucial, and more a "nice-to-have", but for a use case that I _think_ is most common.

I suspect my comments above are because I don't fully understand how the different pieces interact (e.g. the `Store` and `Elasticsearch::Model::Indexing::ClassMethods` modules), so apologies if I'm off the mark with my comments.

estolfo:
> This looks great! :) Appreciate you going to the effort to put this together.

@jits Thanks again for all your feedback! It's really nice of you to take the time to look at the pull requests : )

> Some questions / thoughts:
> 	1.	Given you can pass in options to each new instance, to configure just that instance, do you still need the class level settings (like index_name)? I can see how they provide added flexibility (to configure defaults at the class level), so this isn't a big deal (in my opinion), but just something that consumers need to be aware of (especially as you also have defaults for these class level options, that get used if a value is not provided at all by the consumer, instead of erroring).

There are two main reasons for having the class-level settings. The first is so that you can use the DSL to set up the repository, for example:

```ruby
class NoteRepository
  include Elasticsearch::Persistence::Repository

  document_type 'note'
  index_name 'notes_repo'

  settings number_of_shards: 1, number_of_replicas: 0 do
    mapping dynamic: 'strict' do
      indexes :foo do
      indexes :bar
    end
    indexes :baz
  end
end
```

The second is, `create_index!`, `mappings`, `settings`, etc are implemented in [this module](https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-model/lib/elasticsearch/model/indexing.rb#L83-L304) called `ClassMethods`.  I would actually rather not provide class-level methods and settings and could include the `ClassMethods` module to the Repository mixin so that they are only instance methods; however, then you wouldn’t have any of the DSL methods you see above and you’d have the awkwardness of including `ClassMethods` that are, in fact, instance methods. I’m not sure how important it would be to users to be able to set the options using the DSL. What do you think?

> 	1.	Is there much need for the .with helper? Don't get me wrong, it does seem like a useful helper, but given that it does a shallow .clone it's not certain that you'll get a completely isolated instance. Instead I'd advocate for consumers creating a fresh new instance when they need and managing common settings themselves (as with regular classes and objects).

This is a good point. I provided the `#with` helper to keep instances immutable and to satisfy anyone who would still want a kind of “mutability” of repository instances. It doesn’t matter to me, so for simplicity, we can just take `#with` out and if someone absolutely needs to spawn repository instances from other repository instances, I can add it later.

> 	1.	Alternatively to the above point, if a helper is really needed to instantiate a similar repo, you could have a class level helper method MyRepo.from(my_repo_instance, new_options) which just returns MyRepo.new(my_repo_instance.options.merge(new_options))?
> 	2.	This covers nicely the need for flexibility in creating different – and isolated – repository instances that could talk to different clusters and/or indexes (with the caveat of the comment above); I'd still suggest having a GenericRepository class that can be used to instantiate a very basic repository without having to extend anything. This is certainly not crucial, and more a "nice-to-have", but for a use case that I think is most common.

I’d like to encourage users to create their own Repository classes so that they can be explicit about what settings they want for it. I don’t see a huge benefit of having a generic repository class, when you can just write class `BaseRepository; include Elasticsearch::Persistence::Repository; end`

> I suspect my comments above are because I don't fully understand how the different pieces interact (e.g. the Store and Elasticsearch::Model::Indexing::ClassMethods modules), so apologies if I'm off the mark with my comments.

I think I'll try out some changes discussed above and see how it works out.

estolfo:
@jits Just FYI, I've merged this and released 6.0.0.pre if you'd like to try it out. Thanks!

jits:
@estolfo – awesome! Thanks for the heads up. I'm hoping to upgrade one of our apps to ES 6.x in the foreseeable future, so may get a chance to play with this then.

Issue #774, opened 2018-02-26
Can't Retrieve Records from a Search when Using a Custom ID


roeltm:
When using a custom field as the document id in elasticsearch the basic actions(create/update/delete) are fairly easy to change in the code, e.g.:
```
object.__elasticsearch__.delete_document id: object.custom_id
object.__elasticsearch__.update_document id: object.custom_id
object.__elasticsearch__.index_document id: object.custom_id
```

However, when searching and retrieving the `records`, it seems that the _id is always matched with the primary_key and this is not configurable:
https://github.com/elastic/elasticsearch-rails/blob/b6d485748c71a07d064ea2a46a6da82d64a04cd7/elasticsearch-model/lib/elasticsearch/model/adapters/active_record.rb#L22

https://github.com/elastic/elasticsearch-rails/blob/b6d485748c71a07d064ea2a46a6da82d64a04cd7/elasticsearch-model/lib/elasticsearch/model/response/records.rb#L36-L38

Should this be made configurable? And if so, which part should be configurable, the `primary_key` or the `_id`.

m1neral:
+1

Need API to normalize `_id` before insertion in SQL.
Frequent case, when `_id`'s are `child:123`, `parent:123` etc

spodlecki:
+1 here.

I've had to monkey patch around to make the gem work as expected, which is not optimal for obvious reasons. With ElasticSearch 6.x getting rid of multi-type indices, I've had to opt to do a "custom type field" to keep the number of shards down in our clusters. I first tried to keep IDs as-is, but problems quickly reared their heads when IDs started to conflict (e.g. updating the document would update a different document of the same ID, fetching by ID fetched all ID records, and so on). So I've made the `_id` field a custom value and the `type` field a constant `doc` value.

> note: ElasticSearch's recommendation of `_doc` type is an illegal format says 6.1.1

estolfo:
@roeltm @m1neral @spodlecki 
I have opened [this PR](https://github.com/elastic/elasticsearch-rails/pull/828) as a proposed implementation of your request. Please take a look, try it out in your testing environments, and let me know what you think. Thanks!

soranoba:
+1
What's the PR's status?

PR #828, opened 2018-08-17
[MODEL] Allow custom lookup fields to be specified when fetching records

estolfo:
This pull request presents a way to use a custom lookup field for linking an ES query result to a record in your relational datastore.
In this example:
```ruby
response = ArticleWithRecordLookup.search(query: { match: { title: 'Test1' } })
response.records(record_lookup_field: 'relational_id', elasticsearch_lookup_field: 'ip_addr')
```
The above will use the values in the `relational_id` field of the elasticsearch results in a query against the `ip_addr` field in the relational datastore.

This PR addresses issue https://github.com/elastic/elasticsearch-rails/issues/774

### DIFF
--- a/elasticsearch-model/lib/elasticsearch/model/adapters/active_record.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/adapters/active_record.rb
@@ -19,13 +19,14 @@ module Elasticsearch
           # Returns an `ActiveRecord::Relation` instance
           #
           def records
-            sql_records = klass.where(klass.primary_key => ids)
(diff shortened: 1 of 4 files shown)


m1neral:
Looks good, thank you,

but what about this method: https://github.com/elastic/elasticsearch-rails/blob/b6d485748c71a07d064ea2a46a6da82d64a04cd7/elasticsearch-model/lib/elasticsearch/model/adapters/active_record.rb#L111 (and other adapters)?

Perhaps, should change the`id` field too.

Issue #546, opened 2016-03-10
Sort order disobeyed


vmatekole:
Hi!

I am working with Spree and ElasticSearch gem and have the following code:

``` records = search_result.limit(Spree::Config.products_per_page).page(page).records ```
``` records.joins(:classifications).where("spree_products_taxons.taxon_id IN(#{@taxon.id})").reorder("spree_products_taxons.position asc") ```

Resulting SQL:
```
"SELECT \"spree_products\".* FROM \"spree_products\" INNER JOIN \"spree_products_taxons\" ON \"spree_products_taxons\".\"product_id\" = \"spree_products\".\"id\" WHERE \"spree_products\".\"deleted_at\" IS NULL AND \"spree_products\".\"id\" IN (1760, 1666, 1750, 1672, 1815, 1318, 1361, 1762, 1678, 6392, 1808, 1768, 1669, 1759, 1722, 1965, 1745, 6284, 1740, 1743, 7014, 1735, 1826, 1734, 1732, 1683, 1671, 7428, 1773, 1820, 7321, 1772, 1753, 1766, 1742, 1694, 7319, 2564, 6423, 7337, 1806, 1728, 1727, 1819, 1818, 1739, 1729, 1816) AND (spree_products_taxons.taxon_id IN(247))  ORDER BY spree_products_taxons.position asc"
```

the resulting order disobeys the order by, but strangely enough when I do a records.first or records.pluck the output is in the correct order.

vmatekole:
Hi!

It appears my issue is similar to #217. I have had to result to the same solution, return ids and then do another search with a correct order clause. Any thoughts on this?

karmi:
Hi, where the `reorder` method is coming from? 

There's an integration test for the order here, maybe we need to update that: https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-model/test/integration/active_record_basic_test.rb#L208

karmi:
> (...) strangely enough when I do a records.first or records.pluck the output is in the correct order (...)

The records are loaded only when `.to_a` is called on them (which `.first` etc do implicitly). But I guess your issue is that the `ORDER BY` clause is missing from the final SQL query?

fabiode:
@vmatekole I'm having the EXACT same issue, and when I mean EXACTLY, it's true. I'm trying to use `Spree::Classification` to reorder the Elasticsearch result, but with no success at all.

[...]

PR #831, opened 2018-08-30
[MODEL] Fix sort order on ActiveRecord >= 5

jazzytomato:
This is addressing an issue in #546 

The monkey patching regarding sort order was not removed when calling `order` on ActiveRecord versions >= 5.

The code introduced in https://github.com/elastic/elasticsearch-rails/pull/618 did not change the `order` method, this patch fixes it.

I have added a test that fails without my new code. The problem with testing the order with `.first` is that it applies the sql `LIMIT 1`, which would return a collection of 1 item from the database. But we want to test the order of the collection returned. Therefore, I added a test for the first item of the collection with `[0]` which doesn't limit the results returned from the db.

### DIFF
--- a/elasticsearch-model/lib/elasticsearch/model/adapters/active_record.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/adapters/active_record.rb
@@ -56,7 +56,10 @@ module Elasticsearch
             # Redefine the `to_a` method to the original one
             #
             sql_records.instance_exec do
-              define_singleton_method(:to_a) do
+              ar_records_method_name = :to_a
+              ar_records_method_name = :records if defined?(::ActiveRecord) && ::ActiveRecord::VERSION::MAJOR >= 5
+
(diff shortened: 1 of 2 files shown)


estolfo:
Thanks @jazzytomato !

estolfo:
@jazzytomato If I understand your comment correctly, https://github.com/elastic/elasticsearch-rails/issues/546 is not fully fixed with this PR, right?

jazzytomato:
@estolfo Indeed, there is another issue which is a bit more involved. 

The `.order` method is broken when sent to `ActiveRecord ::Relation` instead of `Elasticsearch::Model::Response::Records`, i.e. this will work 
```ruby
Model.search('abc').order('id desc')
```
but this will not work
```ruby
Model.search('abc').distinct.order('id desc')
```

My comment here explains it in more details: https://github.com/elastic/elasticsearch-rails/issues/546#issuecomment-417268226

PR #832, opened 2018-09-05
[RAILS] Log the ElasticSearch query as json

cbarton:
Using `#inspect` on the hash will inspect the values in the hash which
aren't printed according to the JSON spec.

This will change the output from:

```
{query: {range: {created_at: {gt: Tue, 13 Mar 2018 00:00:00 EET +02:00, lt: Wed, 05 Sep 2018 00:00:00 EEST +03:00}}}}
```

to:

```
{"query": {"range":{"created_at":{"gt":"2018-03-13T00:00:00+02:00","lt":"2018-09-05T00:00:00+03:00"}}}}
```

Which makes it a whole lot easier to copy and paste into Kibana for
testing/profiling.

### DIFF
--- a/elasticsearch-rails/lib/elasticsearch/rails/instrumentation/log_subscriber.rb
+++ b/elasticsearch-rails/lib/elasticsearch/rails/instrumentation/log_subscriber.rb
@@ -28,7 +28,7 @@ module Elasticsearch
 
           payload = event.payload
           name    = "#{payload[:klass]} #{payload[:name]} (#{event.duration.round(1)}ms)"
-          search  = payload[:search].inspect.gsub(/:(\w+)=>/, '\1: ')
+          search  = payload[:search].to_json
 
           debug %Q|  #{color(name, GREEN, true)} #{colorize_logging ? "\e[2m#{search}\e[0m" : search}|
         end


cbarton:
I have signed the CLA if that is what is holding this up.

karmi:
@cbarton, the format of the current `#inspect` is intentional — it's meant for easy "inspection" with a naked eye, where the quotes add unncecessary noise. For a different purpose, like pasting it into Kibana, a regular `#as_json` should be used?

cbarton:
@karmi If that is the intended introspection method, then I can close this PR. We can't use `as_json` because then we would still need to strip out the `=>`, at which point it would be the same as the `to_json` output.

While I agree that there is slightly less cognitive load in reading through the `inspect` output, I'd argue that it's negligible for most values except times/dates (as they are expanded). Speaking for myself personally, I tend to find that reading JSON is pretty easy as my brain kind of extracts away the quotes or if its a large structure, I'd pop it into a prettifier to dig in. The current problem is that you cannot use a prettifier and thus you have to dig through a large blob of curly braces (in our use case our queries that we like to debug can be pretty lengthy). Also, if there is a customer that is having difficultly with a particular search result or I am developing a new query, I have to manually recreate the query to ES or transform the inspected version to JSON.

Again, that's just my 2c, so let me know if you want me to close the PR; either is fine by me!

PR #835, opened 2018-09-13
[MODEL] Ensure that specified ActiveRecord order is not overwritten by Elasticsearch search results order

estolfo:
This pull request resolves the following issues
- https://github.com/elastic/elasticsearch-rails/pull/618
- https://github.com/elastic/elasticsearch-rails/issues/546

And is related to this pull request, which partially resolved the issue.
https://github.com/elastic/elasticsearch-rails/pull/831


estolfo:
@jazzytomato Good catch, I'll test it and see if we can remove that. Thanks for your review!

estolfo:
@jazzytomato I've pushed another commit that removes the interception of `#order`, as it's not necessary with the check on `order_values` in the `#records` method. What do you think?

PR #838, opened 2018-09-18
`respond_to_missing?` to silence Ruby 2.4 warnings

emptyflask:
I ran into a bunch of warnings when using Elastic results with
Ruby 2.4+ and trailblazer/cells, because `#method_missing?` without
`#respond_to_missing?` makes Forwardable think all of the Result
attributes are private methods.

https://robots.thoughtbot.com/always-define-respond-to-missing-when-overriding


emptyflask:
I've signed the CLA.

PR #843, opened 2018-09-24
[RAILS] Fix seeds file to stop using outdated YAML method

shime:
The expert template was using the outdated YAML.load_documents, which is no longer present in the newer versions of Ruby.


estolfo:
Thanks @shime !

PR #845, opened 2018-10-05
Prevent double calls to search.execute! by using cached raw_response

coneybeare:
I was noticing a ton of duplicate search logs in my app, and tracked it down to the `response` method running a non-cached `search.execute!`. An example of when this might occur is the following

````ruby
# Setup search with query and aggregations
response = Post.search('foo')
# Actually execute the search and get results
results = response.results 
# Make a second execution because results raw_response is not being cached
aggregations = response.aggregations
````

By replacing the `search.execute!` within the response method, we can either set, or utilize the cached `raw_results` and prevent double searches.

### DIFF
--- a/elasticsearch-model/lib/elasticsearch/model/response.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/response.rb
@@ -26,7 +26,7 @@ module Elasticsearch
         # @return [Hash]
         #
         def response
-          @response ||= HashWrapper.new(search.execute!)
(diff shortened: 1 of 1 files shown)


estolfo:
Hi @coneybeare 
Thanks for this pull request! It seems that this change has implications for pagination, as made apparent by the test failures. Do you want to take a look at the failures and see if there's an alternate solution that wouldn't affect pagination functionality?

coneybeare:
Will resubmit with a cleaner merge

PR #850, opened 2018-10-15
[MODEL] Avoid executing search twice; Reuse response in Response#raw_response

estolfo:
Alternative to https://github.com/elastic/elasticsearch-rails/pull/847

### DIFF
--- a/elasticsearch-model/lib/elasticsearch/model/response.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/response.rb
@@ -60,7 +60,7 @@ module Elasticsearch
         # Returns the statistics on shards
         #
         def shards
-          @shards ||= HashWrapper.new(raw_response['_shards'])
+          @shards ||= response['_shards']
         end
 
         # Returns a Hashie::Mash of the aggregations
@@ -76,7 +76,7 @@ module Elasticsearch
         end
 
         def raw_response
-          @raw_response ||= search.execute!
+          @raw_response ||= @response ? @response.to_hash : search.execute!
         end
       end
     end


coneybeare:
I think it looks great and would solve my particular issue as well, as long as `@response.to_hash` actually returns the exact same object as a `raw_response` would.

estolfo:
Hi @coneybeare Great! I'm glad it resolves your issue. The `#to_hash` implementation is [here](https://github.com/intridea/hashie/blob/506c385ccc034eaf1e3420ce9940749b3acdc42b/lib/hashie/hash.rb#L18-L39) so it will return the exact hash representation. But perhaps you are referring to whether the keys will be identical. If that's the case, I believe they will be because `HashWrapper` inherits from `Hashie::Mash` and it looks like nothing is done to remove keys when it is initialized [here](https://github.com/intridea/hashie/blob/506c385ccc034eaf1e3420ce9940749b3acdc42b/lib/hashie/mash.rb#L128-L131).

iridakos:
Hi @coneybeare @estolfo 

This implementation **still triggers double searches** if the `raw_response` is accessed before the `response` method.

To overcome this I used the following monkey patch:

```ruby
module Elasticsearch
  module Model
    module Response
      class Response
        def response
          @response ||= HashWrapper.new(raw_response)
        end

        def raw_response
          @raw_response ||= search.execute!
        end
      end
    end
  end
end
```

Referring commit: https://github.com/elastic/elasticsearch-rails/commit/62e2faedd8bf2c68f4762607acdc15ff62768220#commitcomment-33515229

PR #860, opened 2018-11-19
Add license, notice and license headers

estolfo:
Need confirmation that this is the correct way to add the legal information, then will add files and headers to sub-directories.


estolfo:
Closing in favor of https://github.com/elastic/elasticsearch-rails/pull/861

PR #862, opened 2018-12-06
Only execute update if document attributes is not empty

jonbwhite:
Currently if a model has an attribute changed, but that attribute is not part of the elasticsearch template, an empty update will be made to elasticsearch.  This change only follows through on the update if there is content in the distilled attributes hash.


estolfo:
@jonbwhite Thanks for this pull request. Would you mind adding a test for this scenario Thanks!

jonbwhite:
@estolfo added the test!

jonbwhite:
@estolfo rebased and tests are passing now

estolfo:
Thank you @jonbwhite !

PR #881, opened 2019-06-05
Fix Broken Links in Documentation

nodanaonlyzuul:
When Elasticsearch changed their FQDN from `www.elasticsearch.org` to `www.elastic.co`,
they preserved most of their page's paths but didn't setup granular redirect rules.

## For example:

Going to `http://www.elasticsearch.org/contributor-agreement/` will redirect browsers
to https://www.elastic.co/, but drop the path `/contributor-agreement/`.

## Contributing

I have signed the CLA. I haven't found any open issues that reference this problem.
Thanks to all maintainers for their work on this library - let me know if there's any additional work you'd like to see in this PR.


estolfo:
thanks @nodanaonlyzuul !

PR #887, opened 2019-07-03
[Model] Fix naming with inheritance when using Proxy

tmandke:
Closes elastic#717 & Closes elastic#585

### DIFF
--- a/elasticsearch-model/spec/elasticsearch/model/naming_inheritance_spec.rb
+++ b/elasticsearch-model/spec/elasticsearch/model/naming_inheritance_spec.rb
@@ -19,105 +19,185 @@ require 'spec_helper'
 
 describe 'naming inheritance' do
 
-  before(:all) do
-    class ::TestBase
-      extend ActiveModel::Naming
+  context 'without using proxy' do
+    before(:all) do
+      TestBase = Class.new do
+        extend ActiveModel::Naming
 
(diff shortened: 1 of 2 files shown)


elasticsearch-model/spec/elasticsearch/model/naming_inheritance_spec.rb
```diff
@@ -19,105 +19,185 @@
 
 describe 'naming inheritance' do
 
-  before(:all) do
-    class ::TestBase
-      extend ActiveModel::Naming
+  context 'without using proxy' do
```
tmandke: I have just move d the tests into a context and changed the way the classes are defined so that they can be redefined with a different superclass in the other context.

elasticsearch-model/spec/elasticsearch/model/naming_inheritance_spec.rb
```diff
@@ -19,105 +19,185 @@
...
+      end
     end
+  end
+
+  context 'when using proxy' do
+    before(:all) do
+      TestBase = Class.new do
+        extend ActiveModel::Naming
+
+        include Elasticsearch::Model
```
tmandke: I could just copy the required code from `Elasticsearch::Model` in here but that would result in tests passing even if the `Elasticsearch::Model` changed.

estolfo:
thanks @tmandke !

PR #895, opened 2019-07-26
[MODEL] Add warning and documentation about STI support being deprecated

estolfo:


tmandke:
If this option is removed it will mean that if I use STI and want to use the same ES index to store all the subtypes I will need to set the index name on all of them (not sure what else will break at this point).

estolfo:
@tmandke You would have to manually put an artificial type field in each document. You can do this in the `as_indexed_json` method. Then, you'll have to make sure any requests to elasticsearch take into account that "type" field. I would recommend putting different models in different indices. Would that be possible for you?

tmandke:
I am totally for the one "index" for one "type" and we are doing that. 

When using STI we have one base with N variations, and in our case (i bet for a lot of other people too) they have no difference when it comes to search. When we perform a search we want it to be across all the N Variations of the STI model. If I split these N variations across N indexes with exactly the same fields, this is wasteful because some of these types have very few records.

All in all when using STI I feel like things for search are going to be identical/similar enough that it will be beneficial to having all of them in one index because at their core they spawn from the same type.

estolfo:
Are you familiar with the [Multimodel search](https://github.com/elastic/elasticsearch-rails/tree/master/elasticsearch-model#searching-multiple-models)? Maybe that provides what you're looking for?

tmandke:
Yes, we do use the Multimodel search for our autocomplete where we search across all our models.

I am just trying to make a case for supporting the feature to allow models that use STI to share an index. IMO for most cases where STI is used its not worth splitting the generic type into multiple indexes.

estolfo:
Yes, but using a single index for entities with different properties is an anti-pattern and one that Elasticsearch server itself has deprecated and will remove in 8.0. Please see [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html) for more information.
You can always manually implement STI but this library will no longer support it as a feature.

PR #897, opened 2019-08-05
Remove unnecessary `self` as a return value of `initialize` method

pocke:
`initialize` method returned value is not used usually. Because the method is called by `Class.new`.
So the `self` are not necessary.

note
===

If it is called explicitly, the caller can use the returned value. But it is a super rare case.

```ruby
class C
  def initialize
    1
  end
end


c = C.new
p c.__send__ :initialize # => 1
```


estolfo:
thanks!

PR #907, opened 2019-09-05
Small update README for asynchronous Callbacks

shushpan:
Sidekiq requires `perform` method for workers, with `perform_async` it doesn't work.


picandocodigo:
Thanks for this fix @shushpan!

PR #911, opened 2019-10-22
Update: refresh index for persistence update api

rshiva:


elasticcla:
Hi @rshiva, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/elasticsearch-rails/pull/911.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

rshiva:
> Hi @rshiva, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/elasticsearch-rails/pull/911.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

rshiva:
Both emails are already added to my profile.

PR #928, opened 2020-03-27
[DOCS] Move `gem install` to a non-Ruby code block

orhantoy:
`gem install ...` is not Ruby, and the other places with `gem install` also use the indented block style.


picandocodigo:
Thank you @orhantoy!

Issue #924, opened 2020-03-16
Elasticsearch::Model overrides class .search method


Kukunin:
Hello. I've updated from `elasticsearch-model` 6.1.0 version to the latest 7.0.0 and got a silly bug. I have the following structure:

```ruby
class Article
  include Searchable
end
```

and 

```ruby
class Searchable
  extend ActiveSupport::Concern

  included do
    include Elasticsearch::Model
  end

  class_methods do
    def search(term)
      # my logic here
    end
  end
end
```

When I call `Article.search` I expect my method to be called, and it worked such before. With the `elasticsearch-model` 7.0.0 the delegated `Article.__elasticsearch__.search` is called directly. 

I found the reason in the code: https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-model/lib/elasticsearch/model.rb#L115

```ruby
    def self.included(base)
      base.class_eval do
        include Elasticsearch::Model::Proxy

        # Delegate common methods to the `__elasticsearch__` ClassMethodsProxy, unless they are defined already
        class << self
          METHODS.each do |method|
            delegate method, to: :__elasticsearch__ unless self.respond_to?(method)
          end
        end
      end

      # Add to the model to the registry if it's a class (and not in intermediate module)
      Registry.add(base) if base.is_a?(Class)
    end
```

The problem is that `self` from `self.respond_to?` is a singleton class of `Article` class and of course it doesn't respond to search. It's in its public instance methods. 

It was changed 8 months ago here https://github.com/elastic/elasticsearch-rails/commit/f9a87ff3d07a6a95c771ed34f348f71e0f856221#diff-bada22348f9db4f589962054cc9adcb7R115, and I wonder how this was unnoticeable for such a long period. Am I missing something?

The previous implementation was right: `delegate method, to: :__elasticsearch__ unless self.public_instance_methods.include?(method)`

Kukunin:
As a workaround, you can define an empty method for the singleton class of a class before you include Elasticsearch::Model:

```ruby
class Searchable
  extend ActiveSupport::Concern

  included do
    singleton_class.singleton_class.define_method(:search) {}
    include Elasticsearch::Model
  end

  class_methods do
    def search(term)
      # my logic here
    end
  end
end
```

PR #893, opened 2019-07-24
[MODEL] Refactor module/mixin organization for clarity

estolfo:

PR #936, opened 2020-04-22
[MODEL] Do not override existing methods

michaelmcchen:
This PR reverts the line from this commit: https://github.com/elastic/elasticsearch-rails/pull/893/files to address the issue described here https://github.com/elastic/elasticsearch-rails/issues/924

My biggest concern is that this is a very subtle change in behavior that could lead to bugs, especially in a project where custom methods with similar names (import, search, etc.) are defined. It seems safer to keep the existing behavior (pre 7.x.x elasticsearch-rails) and force users who are facing this sort of conflict to be more explicit about their method choice (via `ModelName.__elasticsearch__.method()`)

### DIFF
--- a/elasticsearch-model/spec/elasticsearch/model/module_spec.rb
+++ b/elasticsearch-model/spec/elasticsearch/model/module_spec.rb
@@ -48,6 +48,7 @@ describe Elasticsearch::Model do
       end
 
       DummyIncludingModel.__send__ :include, Elasticsearch::Model
+      DummyIncludingModelWithSearchMethodDefined.__send__ :include, Elasticsearch::Model
     end
 
     after(:all) do
--- a/elasticsearch-model/lib/elasticsearch/model.rb
+++ b/elasticsearch-model/lib/elasticsearch/model.rb
@@ -112,7 +112,7 @@ module Elasticsearch
         # Delegate common methods to the `__elasticsearch__` ClassMethodsProxy, unless they are defined already
         class << self
           METHODS.each do |method|
-            delegate method, to: :__elasticsearch__ unless self.respond_to?(method)
+            delegate method, to: :__elasticsearch__ unless self.public_instance_methods.include?(method)
           end
         end
       end


michaelmcchen:
Agreement review in progress, just wanted to put up this PR in the meantime

michaelmcchen:
A rep from my company signed the CLA (via Docusign), and included my name & Github handle in the company agreement. I think the bot doesn't take that into account, hence the failure 

@picandocodigo (tagging you because you recently released a new version of this gem), do you think you could help me route this PR to the right folks for review? I appreciate it!

picandocodigo:
Hi @michaelmcchen, thanks for this contribution! I'll take a look at it tomorrow :+1:

greenonion:
Hi, we're in the process of changing our codebase so we can upgrade our clusters to ES 7x and we've stumbled on this as well. Can we expect this PR to be merged?

Also to address @michaelmcchen 's concern, I think this behavior should probably have some unit tests to help avoid any regressions :/

picandocodigo:
I'll bump the priority on this issue and let you know about a future release :+1:

greenonion:
@picandocodigo thank you very much! If you'd like we could maybe try to write some tests too!

picandocodigo:
@greenonion definitely, some tests would really help. Thanks!

michaelmcchen:
> Also to address @michaelmcchen 's concern, I think this behavior should probably have some unit tests to help avoid any regressions :/

Agreed! I've updated an existing spec to cover this case and tested it out by reverting my original commit (which now fails the spec).

elasticsearch-model/spec/elasticsearch/model/module_spec.rb
```diff
@@ -48,6 +48,7 @@ def self.search(query, options={})
       end
 
       DummyIncludingModel.__send__ :include, Elasticsearch::Model
+      DummyIncludingModelWithSearchMethodDefined.__send__ :include, Elasticsearch::Model
```
greenonion, line 51: Ah so there was a test for this (line 71) but it wasn't actually set up correctly. Nice!

Issue #967, opened 2020-12-06
Indexer example fails on ES 7+


dixpac:
Indexer example:

```ruby
class Indexer
  include Sidekiq::Worker
  sidekiq_options queue: 'elasticsearch', retry: false

  Logger = Sidekiq.logger.level == Logger::DEBUG ? Sidekiq.logger : nil
  Client = Elasticsearch::Client.new host: 'localhost:9200', logger: Logger

  def perform(operation, record_id)
    logger.debug [operation, "ID: #{record_id}"]

    case operation.to_s
      when /index/
        record = Article.find(record_id)
        Client.index  index: 'articles', type: 'article', id: record.id, body: record.__elasticsearch__.as_indexed_json
      when /delete/
        begin
          Client.delete index: 'articles', type: 'article', id: record_id
        rescue Elasticsearch::Transport::Transport::Errors::NotFound
          logger.debug "Article not found, ID: #{record_id}"
        end
      else raise ArgumentError, "Unknown operation '#{operation}'"
    end
  end
end
```

Doesn't work for me on ES 7.  When I change the type from `article -> _doc` then it works:

```ruby
Client.index  index: 'articles', type: '_doc', id: record.id, body: record.__elasticsearch__.as_indexed_json

#or 
Client.index  index: 'articles' id: record.id, body: record.__elasticsearch__.as_indexed_json
```

-------------

Maybe I'm doing something wrong or this is expected change but didn't yet get into README since 7 branch is still not released. I'm just posting it here so maybe it can help someone.

dixpac:
Opened a PR maybe doesn't make a sense 😅

PR #968, opened 2020-12-06
Fixes `Indexer` README to work with ES 7+

dixpac:
Old Indexer example doesn't work on ES 7. Main problem is specifying
`type` inside `Client.index`.

This PR changes the example to work.

Connected to #967

### DIFF
--- a/elasticsearch-model/README.md
+++ b/elasticsearch-model/README.md
@@ -506,8 +506,8 @@ with a tool like [_Resque_](https://github.com/resque/resque) or [_Sidekiq_](htt
 class Article
   include Elasticsearch::Model
 
-  after_save    { Indexer.perform_async(:index,  self.id) }
-  after_destroy { Indexer.perform_async(:delete, self.id) }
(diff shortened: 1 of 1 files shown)


picandocodigo:
Hi @dixpac,
Thank you for this contribution! Can you sign the Contributor Agreement please?

We may instead want to just stop using `type` at all in the example in the README since it's been deprecated. I'll take a look to update any more places where we're referring to type when using 7.x versions of the stack.

dixpac:
@picandocodigo Just signed, sure why check is still red, maybe there is a latency :)

dixpac:
@picandocodigo no problem, signed!

Quicks:
Hi @dixpac
Maybe also we need to call IndexerWorker after commit changes to database
Cause in postgres with json field I had an issue with that

after_commit  { ElasticSearch::IndexerWorker.perform_async((persisted? ? :index : :delete), self.class.to_s, self.id) }

Issue #980, opened 2021-03-01
Want to stop noisy FATAL `[404]` for index checking on `Elasticsearch::Persistence::Repository`


aeroastro:
## Summary

We are using `Elasticsearch::Persistence::Repository`, and we are happy with the handiness compared with directly using Elasticsearch::Client.

However, we're now concerned about the noisy FATAL error `[404]` the repository emits when checking the index's existence.

* `Elasticsearch::Persistence::Repository#index_exists?`
* `Elasticsearch::Persistence::Repository#create_index!`

## Reproduction Code

elasticsearch-persistence version: 7.1.0 (current master)
ruby version: 2.6.2

```ruby
require 'logger'
require 'elasticsearch/persistence'

class MyRepository
  include Elasticsearch::Persistence::Repository
end

repository = MyRepository.create(
  index_name: 'new_index',
  client: Elasticsearch::Client.new(logger: Logger.new(STDOUT, level: :error))
)
repository.client.indices.delete(index: '*') # initialization for reproduction code

repository.index_exists? # This is the problem
# F, [2021-03-02T00:24:33.962978 #28878] FATAL -- : [404]
# => false
repository.create_index! # This is also the problem
# F, [2021-03-02T00:24:57.883461 #28878] FATAL -- : [404]
# => {"acknowledged"=>true, "shards_acknowledged"=>true, "index"=>"new_index"}
```

## Root Cause

Both of the above codes are using the following method.

https://github.com/elastic/elasticsearch-rails/blob/16a297a56e3d1845cc306c1541794280d4b2fdcc/elasticsearch-model/lib/elasticsearch/model/indexing.rb#L270-L274

Because `Elasticsearch::Client` regards all the HTTP status codes `>= 300` as FATAL, the above code `self.client.indices.exists(index: target_index)` generates FATAL log.

Based on the objective of HEAD request and `index_exists?`, the log is noisy.

## Proposed Solution

Based on the objective of `exists?`, where both `200` and `404` are the expected status code, IMHO, we should avoid fatal logs in these situations.

Now I propose the following two solutions to this issue.

Difficult but fundamental fix:
* Treat 404 for HEAD request as non-fatal in `Elasticsearch::Client`.

As this touches the internal design of `elasticsearch-ruby`, it takes time for the maintainers to decide.

Easy but workaround fix:
* Pass `ignore: 404` as the optional arguments to `Elasticsearch::API::Indices::Actions#exists?`

[...]

PR #981, opened 2021-03-01
Update dependency to explicitly support version 7

aeroastro:
While creating a patch for #980, I have found out the currently supported
`elasticsearch` version by `elasticsearch-model` is unnecessarily wide. ( `> 1` )

Also, the previous dependency of `> 1` and the nonexistence of version-parameterized test
means that the CI is not running against versions under 7.x. for `elasticsearch-model`.

I doubt the current code on `master` still works for old versions,
and I would like to avoid writing code to support compatibility on old versions on master branch.

To solve this issue, I have updated the dependency based on Elasticsearch gems' version policy.
This improves gem consistency and maintenance efficiency.

c.f. 
https://github.com/elastic/elasticsearch-rails/tree/756a4dae1c51331a54bab8f00b0fcb88b4a480ad/elasticsearch-model#compatibility
#875

### DIFF
--- a/elasticsearch-model/elasticsearch-model.gemspec
+++ b/elasticsearch-model/elasticsearch-model.gemspec
@@ -42,7 +42,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version = '>= 2.4'
 
   s.add_dependency 'activesupport', '> 3'
-  s.add_dependency 'elasticsearch', '> 1'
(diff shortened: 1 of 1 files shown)


picandocodigo:
Thanks for this contribution @aeroastro!

You're absolutely right :+1:

aeroastro:
Thank you ❤️

PR #996, opened 2021-07-19
Add max size limit to requests for bulk import

jpr5:
This commit adds a new parameter `max_size`, in bytes, which is used to
enforce an upper limit on the overall HTTP POST size.  This is useful
when trying to maximize bulk import speed by reducing roundtrips to
retrieve and send data.

This is needed for scenarios where there is no control over
Elasticsearch's maximum HTTP request payload size.  For example, AWS'
elasticsearch offering has either a 10MiB or 100MiB HTTP request payload
size limit.

`batch_size` is good for bounding local runtime memory usage, but when
indexing large sets of big objects, it's entirely possible to hit a
service provider's underlying request size limit and biff the import
mid-run.  This is even worse when `force` is true - then the index is
left in an incomplete state with no obvious value to adjust batch_size
down to, in order to sneak under the limit.

The `max_size` defaults to `10_000_000`, to catch the worst-case
scenario on AWS.

### DIFF
--- a/elasticsearch-model/lib/elasticsearch/model/importing.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/importing.rb
@@ -145,6 +145,7 @@ module Elasticsearch
           transform    = options.delete(:transform) || __transform
           pipeline     = options.delete(:pipeline)
           return_value = options.delete(:return)    || 'count'
+          max_size     = options.delete(:max_size)  || 10_000_000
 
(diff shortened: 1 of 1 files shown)


jpr5:
Signed the agreement.

jpr5:
Well, willing to look at/fix the failures but can't see the test detail failures anymore...

PR #1019, opened 2022-02-03
Make the same request independently of the order that the models are loaded

aronwolf90:
Currently, the order of the index in  a search request changes
with the order that the models are loaded.

Problem: This does not matter for real rest request but
makes problems with mocks (E.g. VCR gem) on tests. The order of the tests can change
and in consequence the autoloader from rails loads models in a other order, causing that the
mock does not longer match since the order of indexes in the rest request not longer matches.


aronwolf90:
?

aronwolf90:
Ping

PR #1023, opened 2022-03-20
[MODEL] Added functionality to provide custom value of total count for pagination

bk-az:
### The Problem
When using [collapse parameter](https://www.elastic.co/guide/en/elasticsearch/reference/current/collapse-search-results.html) to collapse search results, the total number of hits in the response indicates the number of matching documents **without collapsing**. As a result, the total number of pages for pagination is miscalculated.

For Example, the following query on the `group_items` index returns the total of 100 results without collapsing.

```sh
GET group_items/_search
{
  "query": {
    "match_all": {}
  }
}
```
But when the search results are collapsed based on the `group_id` field value. The total results returned by the query becomes only 5 instead of 100.

```sh
GET group_items/_search
{
  "query": {
    "match_all": {}
  },
  "collapse": {
    "field": "group_id"         
  }
}
```
The problem occurs when we paginate results, after collapsing, the value of the total number of hits is still 100 in response, as mentioned in the reference:

> The total number of hits in the response indicates the number of matching documents without collapsing. The total number of distinct groups is unknown.

### The solution
This problem with pagination can be solved by providing a custom value of `total_entries` (_will_paginate_) or `total_count` (_kaminari_).  The value of total distinct groups can be calculated by using [cardinality aggregation](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-metrics-cardinality-aggregation.html)

```ruby
# Assuming a method `total_collapsed` is defined to get the total count of 
# collapsed results, using cardinality aggregation.
total_count = total_collapsed(response.aggregations)
```

Now to get pagination working correctly, the value of `total_count` can be specified.

```ruby
# for will_paginate
response.paginate(page: 1, per_page: 30, total_entries: total_count)

# for kaminari
response.page(1).total(total_count)
```

### DIFF
--- a/elasticsearch-model/lib/elasticsearch/model/response/pagination/kaminari.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/response/pagination/kaminari.rb
@@ -106,10 +106,17 @@ module Elasticsearch
             self
           end
 
+          # Set the "total_count" value
+          #
+          def total(value)
+            @total_count = value.to_i
+            self
+          end
+
           # Returns the total number of results
           #
           def total_count
-            results.total
+            @total_count || results.total
           end
 
           # Returns the models's `per_page` value or the default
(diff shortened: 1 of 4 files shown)


bk-az:
@picandocodigo can u please review this PR?

PR #1025, opened 2022-04-07
Add ES 8.x support

ryansch:
In addition to pinning v8 of `elasticsearch-ruby`, there were two other things I had to consider.

ES 8 does not allow sending a document type. I followed previous work in this repo and used `Hash#compact` to remove the `type` param.

`elasticsearch-ruby` v8 uses the new `elastic-transport` gem instead of the older `elasticsearch-transport` one. I added a (hopefully) backwards compatible solution via `ELASTIC_TRANSPORT_CLASS`.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -163,7 +163,7 @@ task :wait_for_green do
         ready = true
         break
       end
-    rescue Elasticsearch::Transport::Transport::Errors::RequestTimeout => ex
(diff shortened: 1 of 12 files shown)


ryansch:
The test suite needs to run with ES 8 on this branch but I'm not familiar enough with it to make the right change.

Issue #1004, opened 2021-08-12
create_index! raises undefined method transport @ 7.2.0


plainlystated:
Upgrading from 7.1.1 -> 7.2.0.  The Changelog mentions "Updates transport references" which sounds related but I couldn't find any further explanation.

I'm hitting this in a spec, on the line:
```
Edition.__elasticsearch__.create_index!(force: true)
```

The stack trace:
```
NoMethodError:
       undefined method `transport' for #<Elasticsearch::Transport::Transport::HTTP::Faraday:0x000055dd1927c0a8>
       Did you mean?  transliterate
     # /usr/local/bundle/gems/elasticsearch-model-7.2.0/lib/elasticsearch/model/indexing.rb:293:in `rescue in delete_index!'
     # /usr/local/bundle/gems/elasticsearch-model-7.2.0/lib/elasticsearch/model/indexing.rb:289:in `delete_index!'
     # /usr/local/bundle/gems/elasticsearch-model-7.2.0/lib/elasticsearch/model/indexing.rb:248:in `create_index!'
     # ./spec/lib/repo/search_spec.rb:6:in `block (2 levels) in <top (required)>'
```

Not sure if I'm doing something wrong or there is a bug related to a recent transport change.

davidwessman:
I am experiencing the same bug.

For me it will only happen when I do `create_index!(force: true)` on an index that does not exist.

- `create_index!` - works
- `create_index!(force: true)` - works if index already exists.

davidwessman:
I tried running the specs locally, in the test for the method that fails I get:

```ruby
client.transport
=>
#<Elasticsearch::Transport::Client:0x00007ffc805e69b8
...
```

when I check in my Rails-code and run:

```ruby
Document.__elasticsearch__.client
=>
#<Elasticsearch::Transport::Client:0x00007f89166235d0
...

Document.__elasticsearch__.client.transport
=>
#<Elasticsearch::Transport::Transport::HTTP::Faraday:0x00007f89166221f8
...
```

So it seems like the test-double for client is setup incorrectly.

sonalkr132:
> I am experiencing the same bug.

[same three](https://github.com/rubygems/rubygems.org/runs/3384049699#step:7:1547) :v: .

It is related to changes introduced in this PR https://github.com/elastic/elasticsearch-rails/pull/999
Updating to elasticsearch 7.14 fixes this error, however now my `rake db:setup` is blowing up with:
```
[...]

PR #1032, opened 2022-06-09
Use self.client instead of client

adibsaad:
Should fix https://github.com/elastic/elasticsearch-rails/issues/1004 (also mentioned here https://github.com/elastic/elasticsearch-ruby/issues/1344#issuecomment-1017176417)

When deleting an index that doesn't exist, the exception handler references `client` but I think it should be referencing `self.client`


Lolote13:
Can someone review this ?

vbrazo (approved):
Please review this PR as it is blocking me

sebfie:
up ?

kalashnikovisme:
I've catch error #1004

Added `gem 'elasticsearch-rails', github: 'adibsaad/elasticsearch-rails', branch: 'patch-1'` instead of main version.

Unfortunately, the error still the same.

PR #1048, opened 2022-11-28
feat: add ES 8.x support

ivamuno:


lancedikson:
@ivamuno Hi! Seems like the tests failed during the run. Could you please update the branch so that the tests would trigger again and we could see into it? I'm also interested in looking into the ES8 support as ES7 comes to the EOL already in a few days.

lancedikson:
The best option would be to rebase the branch on the most recent `main` to make sure it bears all the recent changes as well.

sly7-7:
This could be closed 🎉

PR #992, opened 2021-04-26
Add compatibility with Ruby 3

indirect:
Since Ruby 3 changed the way keyword arguments work, Rails now only accepts keyword arguments for `find_in_batches`. Unfortunately, the `method_missing` implementation on `Elasticsearch::Model::Proxy` doesn't proxy keyword arguments, so Rails now explodes. The error is triggered by calling `Model.import`, and looks something like this:

```
ArgumentError: wrong number of arguments (given 1, expected 0)
bundle/gems/activerecord-6.1.3.1/lib/active_record/relation/batches.rb:128:in 'find_in_batches'
```

The fix is to forward not just positional arguments, but also keyword arguments. This PR adds a test (failing before the patch), and a patch to forward keyword arguments.

Fixes #989, #977.

PR #1053, opened 2023-03-14
Add compatibility with Ruby 3 for 6.x

dersnek:
This is the same change made in https://github.com/elastic/elasticsearch-rails/pull/992, but for version 6.x, to enable support for Ruby 3

All credit for this fix goes to @indirect, I just copy-pasted that code over.


dersnek:
I've read and signed the contributor agreement, but it's not being updated.

shashankjo (approved):

PR #888, opened 2019-07-16
[MODEL] Clear type cache when reloading and ensure model list is unique

estolfo:
See https://github.com/elastic/elasticsearch-rails/issues/848 for issue description

PR #1064, opened 2023-11-10
Addressing #838, STI issues when code is reloaded in development, fix…

### LINKED ISSUE #848: Invalid single-table inheritance type: SubClass is not a subclass of Class
I encountered some serious headaches while working with `Elasticsearch-Model` in development.

My problem is that the `Invalid single-table inheritance type: X is not a subclass of Y` exception was being thrown when I tried to instantiate records with the `Multiple` adapter after my code had been reloaded for the first time. I would have to restart my Rails server every time I made any changes to my code.

What I found was that `Elasticsearch::Model::Registry.models` would endlessly accumulate reloaded classes each time my code was being reloaded. If I had five classes that included `Elasticsearch::Model`, it would increase by 5 each time my code reloaded. I have monkey-patched the `Registry` class with:
```
  module Elasticsearch
    module Model
      class Registry
        def add(klass)
+         existing_model = @models.detect { |model| model.name == klass.name }
+         @models.delete(existing_model)
          @models << klass
        end
      end
    end
  end
```
in an initializer and only for development environment to fix this particular issue to ensure this Registry only contains the newly-reloaded classes.

After this change, I continued to encounter the same error. What I found was that `Elasticsearch::Model::Adapter::Multiple::Records` was caching the result of `__type_for_hit` in a class variable `@@__types`, and after my code was reloaded, the stale classes were being returned instead of the newly reloaded classes. I unset this class variable by hooking into Active Support's reloader:
```
Rails.application.reloader.before_class_unload do
  Elasticsearch::Model::Adapter::Multiple::Records.remove_class_variable(:@@__types) }
end
```
This is in the same environment block that the above monkey-patch is in to avoid affecting production environment. This allowed my records to be instantiated even after my code had been reloaded and fixed my issue.

This issue took a ton of time and effort (and frustration) to debug and fix, it would be nice to see the team provide better support for STI since this is a core feature in Rails. Hopefully this report will help you guys improve on some of these weak spots. I'm happy to provide any other information you might need.

vanboom:
The class level map using @@__types messes up the autoloading when STI classes are present.  The Registry.all.detect lookup should be very fast so this seems to be a good trade off to fix the old STI auto load issue that was reported in #838 and closed by the stalebot.  

Related PR #888 was also closed with a statement that STI support is deprecated and removed from the gem.  It is my opinion that this PR and #888 is a fix related to Rails autoloading and has nothing to do with STI support.   

Please consider this PR to close #848 for good.   Thanks!

### DIFF
--- a/elasticsearch-model/lib/elasticsearch/model/adapters/multiple.rb
+++ b/elasticsearch-model/lib/elasticsearch/model/adapters/multiple.rb
@@ -106,16 +106,16 @@ module Elasticsearch
           # @api private
           #
           def __type_for_hit(hit)
-            @@__types ||= {}
 
             key = "#{hit[:_index]}::#{hit[:_type]}" if hit[:_type] && hit[:_type] != '_doc'
             key = hit[:_index] unless key
 
-            @@__types[key] ||= begin
-              Registry.all.detect do |model|
-                (model.index_name == hit[:_index] && __no_type?(hit)) ||
-                    (model.index_name == hit[:_index] && model.document_type == hit[:_type])
-              end
+            # DVB -- not sure the ramifications of this - but do not memoize the model/klass
+            # I do not think that caching the types is necessary, minimal processing savings
+            # at the expense of broken STI autoloading in development, see #848
+            Registry.all.detect do |model|
+              (model.index_name == hit[:_index] && __no_type?(hit)) ||
+                  (model.index_name == hit[:_index] && model.document_type == hit[:_type])
             end
           end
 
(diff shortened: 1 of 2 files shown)


elasticsearch-model/lib/elasticsearch/model/adapters/multiple.rb
```diff
@@ -106,16 +106,16 @@ def __ids_by_type
           # @api private
           #
           def __type_for_hit(hit)
-            @@__types ||= {}
 
             key = "#{hit[:_index]}::#{hit[:_type]}" if hit[:_type] && hit[:_type] != '_doc'
             key = hit[:_index] unless key
 
-            @@__types[key] ||= begin
```
vanboom, line 114: I put Benchmark.measure around this code and contrary to my expectations, the original hash method was WAY faster.  We will have to re-think this change, maybe the method of storing the class->index_name mapping can be improved to mitigate #838 with performance.

shashankjo (approved):

PR #1070, opened 2024-02-01
Update the opensearch-rails gem (RM-2427)

martijnbolhuis:


technige:
Hi @mauricebolhuis. Thanks for checking in, but I'm not sure that "Use opensearch-ruby instead of elasticsearch gem + rename Elasticsearch module name to OpenSearch" is on our roadmap right now. We would be happy to receive your contributions to the original Elasticsearch projects in future though.

PR #1074, opened 2024-03-15
Remove elastic dependencies

rdsun0301:


technige:
It's OK, I think you got away without anyone noticing.

PR #1056, opened 2023-04-21
WIP: Elasticsearch 8.7.x

woto:
It's draft version of attempt to check what should be done to make elasticsearch-rails work with new elasticsearch-ruby 8.7.x and faraday 2.x. I'm not too deep dived in this gems, but I use it in my pet project. So I decided to try.

This links could help during development:

https://www.elastic.co/blog/moving-from-types-to-typeless-apis-in-elasticsearch-7-0
https://github.com/elastic/elasticsearch-rails/pull/1048
https://github.com/elastic/elasticsearch-rails/pull/1025
https://github.com/elastic/elasticsearch-rails/issues/1049

If you have some thoughts/knowledge how to bring this PR to life faster/better, then let me know.

PR #1075, opened 2024-04-12
Satisfy all tests on 8.x branch

sinisterchipmunk:
I found this conversation ( https://github.com/elastic/elasticsearch-rails/pull/1056 ) which led me to the unreleased 8.x branch, but the tests are currently failing.

This pull request makes as few changes as possible in order to make all unit tests pass when running `rake test:all` from the project root.

The first commit makes some changes that I found necessary in order to connect to Elasticsearch 8.13.2 running in Docker on my local machine. It...
* disables SSL certificate validation so that the self-signed certificates can be used out-of-the-box
* changes the startup script to wait for either yellow or green status, as a single ES node is yellow for me under Docker. According to the ES documentation yellow means, "All primary shards are assigned, but one or more replica shards are unassigned." (https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-health.html) This strikes me as a valid state for a test environment so I included this change in the commit.
* Fixes an apparently broken reference to `CERT_DIR`. Since I'm not sure of the history of this constant, I decided to check if it is available using `defined?(CERT_DIR)` and then only omit the option if the constant is not found, using it as before if it exists.

The second commit actually fixes the tests. There was only one logic change needed.
* The `ClassMethodsProxy` needed to use `extend` on the adapter methods, rather than `include`.

I hope that with passing tests we may see a formal 8.x release soon. I am not clear on what else needs to be done but am available to contribute more if needed.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -38,14 +38,16 @@ def admin_client
 
     if test_suite == 'security'
       transport_options.merge!(:ssl => { verify: false,
-                                         ca_path: CERT_DIR })
+                                         ca_path: defined?(CERT_DIR) ? CERT_DIR : nil
+                                       }.compact)
 
       password = ENV['ELASTIC_PASSWORD']
       user = ENV['ELASTIC_USER'] || 'elastic'
-      url = "https://#{user}:#{password}@#{host}:#{port}"
+      url = "https://#{user}:#{password}@#{host || 'localhost'}:#{port || 9200}"
     else
       url = "http://#{host || 'localhost'}:#{port || 9200}"
     end
+    ENV['ELASTICSEARCH_URL'] ||= url
     Elasticsearch::Client.new(host: url, transport_options: transport_options)
   end
 end
@@ -140,7 +142,7 @@ namespace :test do
   end
 
   desc "Run all tests in all subprojects"
-  task all: :wait_for_green do
+  task all: :wait_for_green_or_yellow do
     subprojects.each do |project|
       puts '-'*80
       sh "cd #{project} && " +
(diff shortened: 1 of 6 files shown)


sinisterchipmunk:
Actually, the problem was more subtle. When the tests were failing, changing from `include` to `extend` did make the tests pass, but broke in a real Rails app.

It turns out the call to `class_eval` was triggering different behavior in test than it was in a real Rails app. In test, `class_eval` was called on an instance of `ClassMethodsProxy`, which overrides `method_missing` and passes the call to `class_eval` into the inheriting model. Thus calling `include` as the original code did (before this PR) was including the adapter's methods into the inheriting model, polluting its namespace, while not defining them in the `ClassMethodsProxy` as intended. The resulting methods in the inheriting model were instance methods, not class methods, and so `method_missing` couldn't find them when `__transform` was called at run-time. Changing `include` to `extend` made `method_missing` proxy into the methods, so the tests passed. However, the behavior was still incorrect, because the model's namespace was unintentionally polluted.

In the real application, on the other hand, ActiveSupport defines `Kernel#class_eval`, so that all objects have this method. But its implementation causes a method to be defined on the singleton class of an instance of an object. So when `class_eval` was called, this didn't go into `ClassMethodsProxy#method_missing` at all, but rather called the ActiveSupport implementation. This defined the methods in the right place (the singleton of the `ClassMethodsProxy` instance), but then `extend` made them class methods, not instance methods as they needed to be, and broke the real Rails app.

So the solution was to revert the previous commit, changing `extend` back to `include`, and then simply ensure ActiveSupport is fully loaded in test with `require "active_support/all"`. This causes the test environment to behave the same as the real Rails app.

Although this was technically all that was needed, I decided to also add a test case to check that the adapter's `__transform` method isn't erroneously placed into the inheriting model's namespace, since that test might have made this easier to debug, and verifies the correct behavior of `ClassMethodsProxy` in the first place.

Ironically, this also means no actual business logic change was needed at all, as the test failures were entirely due to the test environment.

picandocodigo:
@sinisterchipmunk this is great work, thank you very much. Great find on the `class_eval` code, and thanks for adding a test. I've incorporated these changes into `8.x` and I'll work on a pre-release of `elasticsearch-rails` `8.0.0` soon. Thanks!

PR #1083, opened 2024-06-18
Fix rake task for removed type code

### LINKED ISSUE #1082: rake elasticsearch:import:model  - ArgumentError: unknown keyword: :type (ArgumentError)
With the commit "Removing type related code b71d5a9dfb1a5212cbad8b6f9321a1e0a6726579" the type related code was removed, but it is still an option in the rake task, which will fail because the now unknown option "type" is passed down to find_in_batches (instead to be removed in the index function). I will post a pull request to fix that issue.

beatjoerg:
Fixes #1082


picandocodigo (approved):
Thank you @beatjoerg! The failures in the build are not related to the change, it's an installation situation with Ruby 3.1. I'll look into it later, and I'll merge this now.

